PR #2, opened 2026-02-25
Fix web UI bugs: charts, progress bar, and display polish

McDic:
## Summary
- Fix inconsistent background color by enforcing dark theme
- Fix all-in equity chart: diverging count bars, correct hover text via customdata, no horizontal gaps
- Fix duplicate progress bar by separating tournament analysis from hand history computation
- Prevent duplicate equity computations with incrementing compute ID
- Keep chart components mounted on tab switch to preserve state
- Remove unused sync allInEquity.ts (dead code)
- Polish: "Top X%" luck display, shared x-axis, fixedrange, remove zero lines from heatmaps

## Test plan
- [ ] Upload hand history files and verify single progress bar
- [ ] Verify equity count chart: green bars go up, red bars go down from zero
- [ ] Verify equity percentage chart hover shows correct rates (not 100%)
- [ ] Switch tabs and verify progress bar state is preserved
- [ ] Verify dark background on both light and dark OS themes
- [ ] Verify VPIP heatmaps have no black lines on empty cells


McDic:
## Code Review

**Files Reviewed:** 7 (5 modified, 1 deleted, 1 re-export)
**TypeScript compilation:** PASS
**Total Issues:** 9

### By Severity
- CRITICAL: 0
- HIGH: 2 (should fix)
- MEDIUM: 3 (consider fixing)
- LOW/NIT: 4 (optional)

**Recommendation: REQUEST CHANGES** -- Two HIGH severity issues should be addressed before merge.

---

### HIGH: `lastComputedRef` set eagerly means failed computations cannot be retried

**File:** `HandHistoryCharts.tsx:60`

Setting `lastComputedRef.current = currentIds` before the async computation means if the computation **fails** (catch block), the IDs are already marked as "computed". A subsequent re-render with the same data sees `hasNewHands = false` and skips recomputation. The user has no way to retry without uploading different data.

**Fix:** Reset `lastComputedRef` in the catch block:
```tsx
} catch (error) {
  console.error('Chart generation failed:', error)
  lastComputedRef.current = new Set() // Allow retry
  setState(prev => ({ ... }))
}
```

---

### HIGH: Both chart trees stay mounted with `display: none`

**File:** `App.tsx:93-102`

Changing from conditional rendering to `display: none` keeps both chart subtrees (including Plotly instances) mounted at all times, doubling persistent DOM and memory. This is an acceptable tradeoff for preserving progress bar state, but should at minimum have a comment explaining the decision. Ideally, lifting computation state to a parent hook would allow conditional rendering while preserving state.

---

### MEDIUM: Diagonal reference line invisible on dark background

**File:** `allInEquityAsync.ts:452`

The reference line uses `rgba(0,0,0,0.25)` (transparent black), which is nearly invisible on the `#242424` dark background that **this same PR enforces**.

**Fix:**
```tsx
line: { color: 'rgba(255,255,255,0.25)', dash: 'dash' },
```

---

### MEDIUM: Stale computation invalidation does not terminate running Web Workers

**File:** `HandHistoryCharts.tsx:186`

When `computeIdRef` is incremented, stale results are correctly ignored, but the spawned Web Workers continue running and consuming CPU. For large datasets with 8 parallel workers, this could be significant. Consider passing an abort mechanism into `collectAllInDataAsync`.

---

### MEDIUM: Diverging bar Y-axis shows negative tick labels for counts

**File:** `allInEquityAsync.ts:335`

The "Hero Lost" bars use negative `base` values, so the Y-axis shows negative numbers for loss counts. This may confuse users who expect counts to be positive. Consider using `tickvals`/`ticktext` to show absolute values, or add an annotation clarifying the diverging layout.

---

### LOW/NIT (no action required)
- Dark-only theme is intentional and appropriate for a poker analytics tool
- `zeroline: false` additions on heatmaps are clean
- `allInEquity.ts` deletion is complete with no dangling references
- Auto-analysis guard change is correct and well-scoped

---

McDic:
## Review Fixes (6e05bc4)

Addressed the review findings:

| # | Severity | Issue | Resolution |
|---|----------|-------|------------|
| 1 | **HIGH** | Failed computation can't retry | Fixed: `lastComputedRef` reset to empty Set in catch block |
| 2 | **HIGH** | Both chart trees always mounted | Added comment explaining the display:none tradeoff |
| 3 | **MEDIUM** | Diagonal reference line on dark bg | No fix needed: Plotly chart background is white by default |
| 4 | **MEDIUM** | Stale workers keep running | Deferred: workers self-terminate on completion, rare edge case |
| 5 | **MEDIUM** | Negative Y-axis tick labels | Fixed: hidden tick labels since diverging base varies by chop count |

PR #17, opened 2026-05-25
Fix luck score not recalculating on incremental hand-history updates

McDic:
## Problem

When hand histories are loaded incrementally, the All-in Equity luck score was computed only from the **most recently added batch** of hands, while the histogram used **all** cached hands. This caused the displayed luck score to mismatch the underlying data. The score also went stale when every hand was already cached (the recompute block was skipped entirely).

The old code even acknowledged this in a comment:
> `// (simplified: we store the latest, but ideally recalculate from all)`

This re-applies the fix from the closed/never-merged #7, adapted to current `master`.

## Fix

- Add `calculateLuckScore(allInData)` to `allInEquityAsync.ts` — builds a fresh `LuckCalculator` over **all** passed hands and returns the score.
- In `HandHistoryCharts.tsx`: remove the `cachedLuckScore` global and the "store only the latest batch" logic; recompute `luckScore` from the full `allCachedData` set right before chart creation, with a staleness guard.

## Verification

- `tsc --noEmit` — clean
- `npm test` — 44/44 pass
- `npm run lint` — no new errors (pre-existing errors unchanged)

### DIFF
--- a/web/src/components/HandHistoryCharts.tsx
+++ b/web/src/components/HandHistoryCharts.tsx
@@ -33,7 +33,6 @@ export interface HandHistoryChartsRef {
 
 // Global cache for equity results (persists across re-renders)
 const equityCache = new Map<string, AllInHandData>()
-let cachedLuckScore = 0
 
 export const HandHistoryCharts = forwardRef<HandHistoryChartsRef, HandHistoryChartsProps>(
   function HandHistoryCharts({ handHistories }, ref) {
@@ -90,7 +89,7 @@ export const HandHistoryCharts = forwardRef<HandHistoryChartsRef, HandHistoryCha
       try {
         const [
           { getChipHistoriesData, getHandUsageHeatmapsData },
-          { collectAllInDataAsync, createAllInEquityChart },
+          { collectAllInDataAsync, createAllInEquityChart, calculateLuckScore },
         ] = await Promise.all([
           import('../visualization'),
           import('../visualization/handHistory/allInEquityAsync'),
@@ -140,7 +139,7 @@ export const HandHistoryCharts = forwardRef<HandHistoryChartsRef, HandHistoryCha
           }))
 
           // Calculate equity only for uncached hands
-          const { data: newAllInData, luckScore: newLuckScore } = await collectAllInDataAsync(
+          const { data: newAllInData } = await collectAllInDataAsync(
             uncachedHands,
             (current, total) => {
               if (!isStale()) {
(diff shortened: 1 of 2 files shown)


McDic:
## 🔍 Automated Code Review

**Branch:** `fix/luck-score-incremental-update-20260526` vs `master` · **Files reviewed:** 2 · **Recommendation:** COMMENT (mergeable)

### Spec compliance — ✅ PASS
The fix correctly resolves the reported staleness bug. Verified against the Rust core: `LuckCalculator::luck_score` builds a Poisson-Binomial PMF over *all* equities (`equity.rs:706-755`), so the result is order-independent and depends only on the full `(equity, actual)` set. Recomputing from `allCachedData` produces the correct value regardless of batching. The `cachedLuckScore` global and the stale "store only the latest batch" block are removed, and the now-unused `luckScore` from `collectAllInDataAsync` is correctly dropped.

### Findings

**[HIGH-1] Redundant luck-score computation in `collectAllInDataAsync`** — `allInEquityAsync.ts:218-237`
`collectAllInDataAsync` still computes a `luckScore` over the uncached batch (full WASM init + loop + `luckScore()`), but the caller now discards it. Every incremental update pays for the computation twice (new batch, thrown away + full set in `calculateLuckScore`). The new function is essentially a copy of lines 219-235. Fix: have `collectAllInDataAsync` stop computing/returning `luckScore`, and extract the shared body.

**[MEDIUM-1] Luck score recomputed from ALL hands on every update (no incremental caching)** — `HandHistoryCharts.tsx:174-175`
Computation is O(n log n) per update via FFT-based Poisson-Binomial (`equity.rs:653-733`), n = total all-in hands. For typical datasets (hundreds to low thousands of all-ins) this is milliseconds — acceptable. Missed opportunity: could memoize on result-set size + win-count to skip recompute when unchanged. Low priority unless profiling shows otherwise. Dynamic `import()`/`default()` per call is fine (module-cached, idempotent).

**[MEDIUM-2] Stale-guard placement** — `HandHistoryCharts.tsx:175-176`
`isStale()` is correctly placed *after* the `await`, preventing stale `setState`. `allCachedData` is a materialized snapshot (lines 166-172) and concurrent runs only *add* to `equityCache`, so no concurrent-mutation hazard. No action required.

**[MEDIUM-3] Silent error swallowing** — `allInEquityAsync.ts:251-264`
Both `addResult` and `luckScore()` errors are caught and ignored (`catch {}`). A failure renders as a legitimate-looking "Luck Score: 0.00 (Top 50.0%)", indistinguishable from a true neutral score. Add a `console.warn` in the catch blocks. (Mirrors a pre-existing pattern, not a regression.)

**[LOW-1] DRY violation** — `allInEquityAsync.ts:246-266`
`calculateLuckScore` is a verbatim duplicate of lines 219-235. Extract a private helper (e.g. `computeLuckScoreFromData`) and call from both sites — also resolves HIGH-1.

**[LOW-2] Empty-data edge case** — `allInEquityAsync.ts:247`
`if (allInData.length === 0) return 0` is correct and matches the empty-data chart path (no luck annotation shown). No issue.

### Suggested follow-up
Address **HIGH-1 + LOW-1** together (extract a shared helper, drop the unused computation in `collectAllInDataAsync`) and add a `console.warn` for **MEDIUM-3**. No CRITICAL/HIGH correctness or security defects — the fix is behaviorally correct and mergeable.

---
🤖 Posted by Claude Code code-reviewer agent

PR #1, opened 2026-02-15
Web Migration: TypeScript + React + WASM implementation

McDic:
## Summary

Complete web migration of pokercraft-local from Python GUI to TypeScript + React + WebAssembly.

### Key Features
- **React + TypeScript frontend** with Vite build system
- **WASM module** compiled from Rust core for equity calculations
- **Web Workers** for non-blocking heavy computations
- **Plotly.js charts** matching Python implementation

### Charts Implemented
- **Tournament Charts**: Historical Performance, RRE Distribution, Prize Distribution, RR by Rank, Bankroll Analysis
- **Hand History Charts**: Chip Histories, Hand Usage Heatmaps, All-in Equity Analysis

### Performance Optimizations
- Parallel equity calculations using multiple Web Workers
- Preflop cache for instant 2-player preflop equity lookups
- Async chart generation with browser yielding
- Data deduplication on file re-upload

### Deployment
- GitHub Pages workflow added (triggers on web/ or Rust changes)
- Deploys to https://mcdic.github.io/pokercraft-local/

## Test plan
- [ ] Verify build succeeds in CI
- [ ] Test file upload with tournament summaries
- [ ] Test file upload with hand histories
- [ ] Verify all charts render correctly
- [ ] Test GitHub Pages deployment

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #18, opened 2026-05-26
Scope HTML export to active tab; remove deprecated demo links

McDic:
## Summary

- **README**: removed the two **Demo** links that pointed to output from the deprecated Python version of Pokercraft Local — they no longer reflect the web app.
- **HTML export** now exports **only the active tab** (tournament *or* hand history) instead of bundling both into one file, with a tab-specific filename.
- Export stays **eager** (button enabled as soon as data loads). When the active tab is still computing, a **themed in-page warning modal** (replacing the blocking native `window.confirm`) tells the user some charts (e.g. All-In Equity) may be missing, and lets them proceed or cancel.
- Export filenames now include time (`YYYY-MM-DD_HH-MM-SS`, colon-free) so same-day exports don't collide.
- Fixed: the warning now also appears while the **bankroll** simulation worker is still running — it's computed by the analysis Web Worker (App-level `isLoading`), not inside `TournamentCharts`, so it was previously omitted with no warning.

Plotly is still loaded via CDN in the exported HTML (no bundling), and output remains a standalone HTML file — both already in place before this change.

## Commits

- Remove deprecated demo links; scope HTML export to active tab
- Add time to export filename; replace native confirm with themed modal
- Show export warning while bankroll worker is still running

## Verification

- `npm run build` ✓
- `npm test` ✓ (44/44)
- `eslint` clean on changed files


McDic:
## 🤖 Code Review (automated, 1 iteration)

**Verdict: COMMENT** — no CRITICAL/HIGH issues; safe to merge. Spec compliance passed. `useImperativeHandle` has no deps array so `isComputing()`/`getChartData()` read fresh state (no stale-closure bug). Findings below are improvements worth considering.

### MEDIUM

**1. Modal has no Escape-key handling** — `web/src/components/ConfirmModal.tsx`
A `role="dialog" aria-modal="true"` element should be dismissable via Escape. Overlay-click cancels, but keyboard users have no equivalent. Add a `keydown` listener for Escape → `onCancel` while `open`.

**2. Modal doesn't move/trap focus; missing accessible name binding** — `web/src/components/ConfirmModal.tsx`
With `aria-modal="true"`, focus should move into the dialog on open (currently stays on the trigger button). Also reference the title via `aria-labelledby` and message via `aria-describedby`. Fix: `tabIndex={-1}` + ref, `.focus()` in an `open` effect, add `id`s + ARIA bindings.

**3. `isLoading` warns on the wrong tab** — `web/src/App.tsx` (`handleExport`)
`isLoading` is a single global flag set by both `parseFiles` and `runAnalysis`, but `runAnalysis` only runs the tournament bankroll worker. Gating both tabs on it means exporting the hand-history tab during a tournament analysis shows a false "still calculating" warning (and vice-versa during parse). Suggestion: `(isTournament && isLoading) || activeRef?.isComputing()`, or thread a tab-specific loading flag.

### LOW

**4. Export button tooltip still says "all charts"** — `web/src/components/Header.tsx:25` (`title="Export all charts as HTML"`). Behavior is now active-tab-only; update the tooltip.

**5. Captured `charts` snapshot can go stale between warning and confirm** — `web/src/App.tsx`. If the user leaves the modal open until computation finishes, "Export anyway" exports the older snapshot. Re-reading `activeRef.getChartData()` at confirm time would give the freshest data (or add a comment if intentional).

**6. Confirm/Cancel buttons aren't autofocused** — `web/src/components/ConfirmModal.tsx`. Enter does nothing useful and tab order starts outside the dialog. Pairs with finding #2; autofocus Cancel (safer default).

**7. `exportTimestamp` uses local time without indication** — `web/src/App.tsx`. Intentional change from UTC `toISOString`; arguably better for users, but exports across timezones sort ambiguously. Non-blocking.

**8. Duplicate `isComputing()`/`getChartData()` ref shape in two components** — `TournamentCharts.tsx` / `HandHistoryCharts.tsx`. A shared `ChartsRef` interface would reduce drift. Maintainability nit.

---
Most worthwhile before merge: the two accessibility gaps (#1, #2) and the cross-tab false-positive (#3).

McDic:
## Review follow-up (commit 23c0308)

**Resolved:**
- **MEDIUM-1** — `ConfirmModal` dismisses on Escape (listener active only while open).
- **MEDIUM-2** — Focus moves to Cancel on open; dialog binds `aria-labelledby`/`aria-describedby` via `useId`.
- **MEDIUM-3** — `isLoading` warning gate scoped to the tournament tab (`(isTournament && isLoading) || activeRef?.isComputing()`), so exporting hand history during a background analysis no longer shows a false warning while still covering the bankroll case.
- **LOW-4** — Export button tooltip now says "active tab's charts" instead of "all charts".

**Deferred (intentionally out of scope):**
- **LOW-5** — `charts` snapshot can go stale between warning and confirm.
- **LOW-6** — beyond the Cancel autofocus added for MEDIUM-2, no further default-action wiring.
- **LOW-7** — `exportTimestamp` uses local time (intentional change from UTC).
- **LOW-8** — duplicate ref shape across the two chart components; a shared `ChartsRef` interface could dedupe later.

Verified: build ✓, 44/44 tests ✓, lint clean.

PR #4, opened 2026-02-25
Remove all Python code after web migration

McDic:
## Summary
- Remove ~6900 lines of dead Python code (source, tests, config, CI workflows, PyO3 bindings)
- Replace `hu_preflop_cache.bin` symlink with actual file in `web/public/`
- Strip all `#[cfg(feature = "python")]` conditional compilation from Rust core crate
- Update CI workflows, pre-commit config, README, and .gitignore
- Bump version to 3.1.1

## Test plan
- [x] `cargo build` passes
- [x] `cargo test --release` passes (8/8 tests)
- [x] All pre-commit hooks pass
- [x] `web/public/hu_preflop_cache.bin` is a real file (not symlink)
- [ ] Verify deploy-web workflow works on merge

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #25, opened 2026-06-21
Fix blank production page: react-plotly.js interop broke under vite 8

McDic:
## Problem

The deployed site rendered a **full black page** whenever data was loaded — both the Tournament and Hand History tabs. It worked in local `npm run dev` and even in a local `npm run preview`, which made it look data-dependent. The only console signal was a minified `Uncaught Error: Minified React error #130` (element type is invalid … got: object).

## Root cause

A **vite 8 build regression** introduced by the dependabot bump in #24 (vite `^7.3.2` → `^8.0.16`).

Vite 8 bundles `react-plotly.js` through esbuild's Node interop, emitting `__toESM(require_react_plotly(), 1)`. With `isNodeMode = 1`, `__toESM` ignores `__esModule` and sets `.default` to the **entire module object**. Since react-plotly's `module.exports` is `{ __esModule: true, default: PlotComponent }`, `import Plot from 'react-plotly.js'` resolved to an object instead of the component, so `<Plot>` threw React #130 and unmounted the whole tree → blank page (the dark `#242424` background).

It only affected production builds: the dev server and vite 7 (Rollup's `getDefaultExportFromCjs`, which honors `__esModule`) resolved the default correctly. The bug was masked locally because `node_modules` still had vite 7 from before the bump, while CI builds with vite 8 from the lockfile.

## Fix

- **`web/src/components/plot.ts`** (new): interop-safe resolver that unwraps `.default` until it reaches the component (a function/class). Correct under both vite 7 and vite 8. Validated against esbuild's exact `__toESM(mod, 1)` semantics (old path → object, new path → the component function).
- `TournamentCharts.tsx` / `HandHistoryCharts.tsx` import `Plot` from `./plot`.
- **`web/src/components/ErrorBoundary.tsx`** (new), wired top-level (`main.tsx`) and per-chart (`App.tsx`): a single failing chart now shows an error message + component stack instead of blanking the entire app. (This is also what pinpointed the failing component during debugging.)

## Verification

- `npm run build` (tsc + vite 8) passes.
- The broken `jsx(import_react_plotly.default, …)` pattern is gone from the bundle; `var Plot = resolvePlot()` is present.
- A vite-8 local build reproduced the deployed artifact (byte-identical chunk hashes), and `npm run preview` with the fix renders all charts on both tabs.


McDic:
## 🤖 Automated code review

_Posted by a Claude Code reviewer agent. Recommendation: **APPROVE** (no blocking issues)._

Solid, well-targeted fix. Verified against the actual installed module: `react-plotly.js@2.6.0`'s main entry is a Babel CJS module with `__esModule = true` and `default = PlotComponent` (a transpiled class, `typeof === 'function'`). The resolver correctly unwraps to that class under both bundlers, `tsc` passes, and the ErrorBoundary is a correct React class boundary.

### Should-fix
**1. ErrorBoundary never resets — a transient error is permanently sticky** (`ErrorBoundary.tsx:24-54`)
Once `state.error` is set, there's no path back to rendering `children`. For the per-chart boundaries (`App.tsx:139/149`), a one-off chart throw leaves the user stuck on the fallback until a full reload, even after re-uploading good data. Consider a `resetKeys` pattern or a "Try again" button calling `this.setState({ error: null, componentStack: null })`. Since wrapped components persist across tab switches, this is a real UX dead-end.

**2. Fallback has no defined styling / dead class name** (`ErrorBoundary.tsx:41`)
The `error-boundary` class has no CSS rule anywhere. On a top-level crash (`main.tsx`), the fallback renders outside `.app` with only the bare body background. Add a minimal rule (background, max-width, color) or drop the unused class.

### Nit
**3. Resolver: add a callable assertion so future bumps fail loudly** (`plot.ts:27`)
The `?? ReactPlotly` fallback is effectively unreachable, but if a future dependency change defeats the loop, it could silently return an object again. Asserting the result is a function would turn a silent re-regression into a clear error:
```ts
const resolved = resolvePlot()
if (typeof resolved !== 'function') {
  throw new Error('react-plotly.js default export did not resolve to a component')
}
```

**4. No test for the resolver** (`plot.ts`)
A tiny unit test (mock the vite-8 double-wrap shape, assert `resolvePlot()` returns the function) would lock in regression protection — valuable since this bug is invisible in dev and only surfaces in production builds.

### Praise
- The root-cause comment in `plot.ts:1-21` is excellent and accurate (`__toESM(..., 1)` / `isNodeMode`, vite-7-vs-8 divergence).
- Defense-in-depth is the right call: resolver fixes the actual bug; ErrorBoundary turns any future React #130-class failure from a silent black page into a diagnosable message with component stack.

### Scope check (other CJS imports)
Grepped for other default-imported CJS packages used as components. The only other third-party default import is `import JSZip from 'jszip'` (`fileLoader.ts:6`), used as `JSZip.loadAsync(...)` — not rendered as JSX, so it can't hit React #130. The fix is correctly scoped to the one package that is both CJS-`__esModule` and rendered as a component.

McDic:
## 🤖 Follow-up: review items addressed (recap)

Applied in `958b6ac`. Recorded here so we can recap later why each choice was made. All 49 unit tests pass and the vite-8 production build is green.

### ✅ Done
1. **Resolver unit tests** (nit #4) — extracted the resolver into `components/cjsInterop.ts` (kept free of the browser-only `plotly` import so it's testable in Node) and added `cjsInterop.test.ts`: function passthrough, single `{default}` unwrap, the vite-8 `__toESM(mod, 1)` double-wrap, the over-unwrap guard, and the no-component case.
2. **Resolver hardening** (nit #3) — when the import doesn't resolve to a function, it now logs an actionable message (`resolveDefaultExport: expected a component/function …`) instead of silently returning an object that would resurface as a cryptic React #130.
3. **ErrorBoundary reset** (should-fix #1) — added a `resetKeys` prop; `App.tsx` passes the loaded datasets, so replacing/re-uploading data clears a stuck fallback without a full page reload.
4. **Fallback styling** (should-fix #2) — added a `.error-boundary` CSS rule matching the existing `.errors` card (the class was previously undefined), and dropped the now-redundant inline padding.

### ⚖️ Done, but differently than suggested
- **Hardening logs instead of throwing.** The review suggested `throw`. A throw happens at module-eval time, which the ErrorBoundary can't catch — it would hard-blank the page, the exact failure mode this PR removes. Logging keeps the actionable signal *and* lets the boundary render a graceful fallback. Net diagnostic value is the same; resilience is strictly better.

### ❌ Deliberately not done (with reasons)
- **No "Try again" button on the ErrorBoundary.** Chart errors here are deterministic for a given dataset, so re-rendering the same children without a data change would just re-throw. `resetKeys` (auto-recover when data changes) is the meaningful trigger; a button would be a no-op in the common case.
- **No `resetKeys` on the top-level boundary** (`main.tsx`). It's the catch-all backstop with no data input to key on; resetting it on a reload isn't applicable.
- **Did not revert the vite 8 bump / pin vite back to 7.** The resolver fixes the interop under vite 8, so staying current with the lockfile/CI is correct — no need to fight dependabot or freeze the toolchain.
- **No broader interop refactor.** The scope check confirmed `react-plotly.js` is the only CJS default rendered as a component (the other CJS default, `jszip`, is called, not rendered). A generic interop layer would be speculative; `resolveDefaultExport` is reusable if another such dep appears.

PR #27, opened 2026-07-12
Fix stale bankroll chart after the first upload

McDic:
## Problem

Bankroll analysis appeared to run only for the first batch of tournament results. The simulation itself *was* re-running on every upload — but `TournamentCharts` never redrew the chart with the new results, so the chart kept showing the numbers computed from the first upload.

## Root cause

The chart-regeneration effect skipped work by comparing input **lengths**:

```ts
if (tournaments.length === lastTournamentCountRef.current &&
    bankrollResults.length === lastBankrollCountRef.current) return
```

A bankroll run always returns one result per initial capital (10/20/50/100/200/500 → 6 entries), so its length is identical even when every number in it changed. On a second upload:

1. The parsed tournaments land first → the guard passes (tournament count grew) → charts regenerate, but the new simulation is still running, so the bankroll chart is drawn from the **previous** results and `lastBankrollCountRef` is recorded as 6.
2. The fresh simulation arrives → effect fires → guard sees `6 === 6` plus the already-recorded tournament count → **early return**. The new results are silently dropped.

The same guard also allowed two chart generations to race: the effect cleanup set the shared `abortRef`, and the immediate re-run reset it to `false`, so a superseded generation could finish last and win with its stale bankroll data.

## Fix

- **`TournamentCharts.tsx`** — compare inputs by identity instead of by length, and supersede in-flight generations with a monotonic compute id (`isStale()`), the pattern `HandHistoryCharts` already uses.
- **`useAnalysisWorker.ts`** — `mergeTournaments` / `mergeHandHistories` now return the *existing* array when a parse adds nothing new, which makes array identity a truthful "the data changed" signal. Sorting moved into the merge, dropping an in-place `.sort()` of React state.
- **`App.tsx`** — the auto-run effect used the same fragile count comparison; it now keys on the tournaments array identity.

## Tests

- `TournamentCharts.test.tsx` replays the real upload sequence (parse #2 lands → charts redraw with stale bankroll → fresh simulation arrives) and asserts the chart is drawn from the newest results. Verified that it **fails against the pre-fix component** and passes with the fix.
- `useAnalysisWorker.test.ts` covers the merge dedup/sort behaviour and the identity-preservation contract the fix depends on.
- Adds `jsdom` as a devDependency — there was no DOM test environment, so React effect-ordering bugs like this one were previously untestable.

Full suite: 57 passing. `npm run build` clean; lint unchanged from the `master` baseline.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -26,11 +26,63 @@
         "eslint-plugin-react-hooks": "^7.0.1",
         "eslint-plugin-react-refresh": "^0.4.24",
         "globals": "^16.5.0",
+        "jsdom": "^29.1.1",
         "typescript": "~5.9.3",
         "typescript-eslint": "^8.48.0",
         "vite": "^8.0.16",
         "vitest": "^4.1.0"
       }
+    },
+    "node_modules/@asamuzakjp/css-color": {
+      "version": "5.1.11",
+      "resolved": "https://registry.npmjs.org/@asamuzakjp/css-color/-/css-color-5.1.11.tgz",
+      "integrity": "sha512-KVw6qIiCTUQhByfTd78h2yD1/00waTmm9uy/R7Ck/ctUyAPj+AEDLkQIdJW0T8+qGgj3j5bpNKK7Q3G+LedJWg==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@asamuzakjp/generational-cache": "^1.0.1",
+        "@csstools/css-calc": "^3.2.0",
+        "@csstools/css-color-parser": "^4.1.0",
+        "@csstools/css-parser-algorithms": "^4.0.0",
+        "@csstools/css-tokenizer": "^4.0.0"
+      },
+      "engines": {
+        "node": "^20.19.0 || ^22.12.0 || >=24.0.0"
+      }
+    },
+    "node_modules/@asamuzakjp/dom-selector": {
+      "version": "7.1.1",
+      "resolved": "https://registry.npmjs.org/@asamuzakjp/dom-selector/-/dom-selector-7.1.1.tgz",
+      "integrity": "sha512-67RZDnYRc8H/8MLDgQCDE//zoqVFwajkepHZgmXrbwybzXOEwOWGPYGmALYl9J2DOLfFPPs6kKCqmbzV895hTQ==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@asamuzakjp/generational-cache": "^1.0.1",
+        "@asamuzakjp/nwsapi": "^2.3.9",
+        "bidi-js": "^1.0.3",
+        "css-tree": "^3.2.1",
+        "is-potential-custom-element-name": "^1.0.1"
+      },
+      "engines": {
+        "node": "^20.19.0 || ^22.12.0 || >=24.0.0"
+      }
+    },
+    "node_modules/@asamuzakjp/generational-cache": {
+      "version": "1.0.1",
+      "resolved": "https://registry.npmjs.org/@asamuzakjp/generational-cache/-/generational-cache-1.0.1.tgz",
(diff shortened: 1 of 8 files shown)


McDic:
## Code review iteration 1

Reviewed the branch with a code-reviewer agent. It found one real defect that I had missed, plus dead code and test-quality issues. Pushed as `aee40e0`.

### Fixed

**Issue 1 — [HIGH] A superseded chart generation could still report "Complete".** With no bankroll results yet — which is exactly the first-upload flow, where the charts are drawn while the simulation is still running in the worker — nothing checked `isStale()` between the last `await yieldToBrowser()` and the terminal `setState`. So a generation that had already been superseded mid-flight would fall through and clear `isComputing` / set the bar to "Complete" while the *newest* generation was still building charts. Two consequences: the progress bar lies, and `App.tsx`'s `activeRef?.isComputing()` export guard is bypassed, so hitting Export in that window silently exports the stale bankroll chart with no "still calculating" warning — the same class of stale-data hole this branch exists to close. I confirmed it independently before patching.

Fixed by guarding the terminal `setState` and the `catch` with `isStale()`, and by superseding the in-flight generation from the effect cleanup (`computeIdRef.current++`), which also matches `HandHistoryCharts` and stops an unmounted component from finishing an expensive generation.

**Issue 2 — [MEDIUM] `lastComputedRef` was unreachable dead code.** The effect's dependency list already compares `tournaments` / `bankrollResults` by `Object.is`, so the identity guard inside the effect could never fire. It was also a latent trap: unlike `HandHistoryCharts`, the `catch` path never reset it, so a failed generation would have wedged permanently if a future change ever made it reachable. Deleted, with a comment on the dep list recording why a count-based skip must never come back. The test that "covered" it was a tautology (it passed on master too) and is gone.

**Issue 3 — [LOW] The tests were timing-dependent.** They polled real timers and leaned on jsdom's timer-driven `requestAnimationFrame`, which drifts under CPU contention on a loaded CI runner. Worse, the original supersede test did not actually reproduce Issue 1 — it re-rendered too early, so the first generation exited at an earlier `isStale()` check and never reached the terminal `setState`. Reproducing it needs the generation parked at its *final* yield when the new results land, which wall-clock polling cannot hit reliably.

Rewrote both tests to mock `yieldToBrowser` into gates the test opens by hand. The interleaving of the two concurrent generations is now exact, with no wall-clock waits anywhere. Verified: the supersede test fails without the Issue 1 guard (`expected false to be true`, 18ms), and both tests fail against the pre-fix component.

**Issue 4 — [LOW] `as HandHistory` cast in a fixture** let it omit required fields (`communityCards`, `actionsPreflop`, …), so it would keep compiling if the merge ever started reading one. Fixtures are now complete, uncast, and shared from `src/test/fixtures.ts` instead of duplicated across the two test files.

### Not changed, with reasons

- **`isComputing` can remain `true` after `reset()`** — the effect cleanup supersedes the in-flight generation, and the effect body then returns early at `tournaments.length === 0`, so nothing clears the flag. It is unobservable: with zero tournaments the component renders the "No tournament data loaded" branch, so there is no progress bar, and the export button is not wired to this ref in that state. `reset()` is also currently unused. Clearing it in the effect body would mean a synchronous `setState` inside a `useEffect`, which trips the `react-hooks` cascading-render rule (an error, not a warning, in this config). `HandHistoryCharts` has the identical shape, so I left them consistent.
- **The `react-hooks/exhaustive-deps` ref warning on the new cleanup** is suppressed with a reason rather than "fixed". Reading the ref's latest value at cleanup is the intent, not a stale-capture mistake; the rule's advice to copy it into a variable would break the mechanism. Lint is back at the master baseline (7 errors, 1 warning — all pre-existing).

### Verification

57 tests passing, `npm run build` clean, `npm run lint` unchanged from master.

### Worth noting (no change needed)

The reviewer spotted that the merge change silently fixes a second latent bug in `HandHistoryCharts`, which this branch does not touch. On master, `mergeHandHistories` returned a fresh array on *every* parse result, so uploading a tournament-only or duplicate file while an equity computation was in flight re-fired that component's effect; its cleanup invalidated the in-flight compute, the body then bailed out via `!hasNewHands`, and nothing ever restarted it — leaving `isComputing` stuck `true` with a frozen progress bar. Preserving array identity means that re-fire no longer happens.

PR #30, opened 2026-07-12
Add English/Korean i18n for the site, charts, and HTML export

McDic:
Adds English + Korean across all three surfaces: the website chrome, the Plotly charts, and the standalone HTML export.

Translations live in `web/src/i18n/locales/<lang>.json` as **flat, dot-separated keys** (Minecraft resource-pack style), so a native speaker can contribute a language by editing one JSON file and never touching TypeScript. See the [translator's guide](web/src/i18n/locales/README.md).

The Korean is **ported from the EN/KO translations the deleted Python app carried** (`git show 89ba853^:pokercraft_local/translation_values.json`) rather than newly written — `순수익`, `멘징 이상 벌은 비율`, `딥런: 32배 수익`, `히어로 패배`. That file also deliberately left poker jargon (`RR`, `RRE`, `PERR`, `VPIP`, `UTG`, `BTN`) in English, which is the policy this PR keeps.

## How it fits together

- **Chart builders take an explicit `t` parameter** instead of importing the i18n singleton, so they stay pure and testable. The chart tests now assert on the translation *key* rather than English prose — a stronger assertion than what they had.
- **The worker stopped speaking English.** It used to post pre-rendered progress text (`"Calculating equity: 3/17"`) across `postMessage`. It now sends a key + params and the main thread renders it, so the worker needs no locale and the progress bar follows a language switch *mid-analysis*.
- **The export was nearly free.** `generateExportHTML` serializes each chart's layout wholesale, so translating the builders translated the export too; only its own chrome, `lang` attribute, and timestamp needed threading. Chart names are now emitted as `<h3>` headings — they were populated but never rendered.
- **Keys are compile-checked.** An i18next module augmentation types `t()` against `en.json`, so `t('header.exprot')` is a TypeScript error. `test-web.yml` gains a typecheck step, because that check previously ran only in the deploy workflow on master — too late for exactly this class of break.

## Two bugs found while verifying

- **A language switch during analysis threw away all in-flight equity work.** The `equityCache` write sat *after* the staleness check, so switching mid-run discarded every hand already computed and made the replacement run recompute the whole WASM batch from an empty cache. Confirmed by A/B in a real browser: reverted, the switch re-runs `0/24 → 24/24`; fixed, it recomputes nothing.
- **A tournament named `</script>...` broke the export.** Pre-existing. `JSON.stringify` escapes neither `<` nor `/`, so a name copied verbatim from an uploaded summary file closed the export's inline `<script>` early, producing a chart-less download. Now escaped, with a regression test that fails without the fix.

## Verification

- `npm run build` (typecheck proves every `t()` key exists), `npm test` — 66 passing.
- Driven end-to-end in a real browser with 120 synthetic tournament summaries and 14 hand-history files: all 8 charts rebuild in Korean (titles, axis titles, legends, hover templates, annotations), the language persists across reload, and a Korean export renders standalone with zero English leftovers.
- `locales.test.ts` rejects unknown keys and placeholder/Plotly-field drift, and reports Korean coverage (currently 100%) **without failing on gaps** — a partial translation is a valid contribution, since i18next falls back to English.

## Not in scope

The review surfaced two **pre-existing** `rrByRank.ts` bugs unrelated to i18n, left for a follow-up: the ITM-cut annotation passes a raw `x: 1/8` on a log axis (Plotly wants log10, so the label lands off-axis at ~133%), and an all-losses upload leaves `minPercentile === Infinity`, yielding `range: [0, Infinity]` and a blank plot.

Worker/parser error text stays English — it is diagnostic and carries file names and raw values.


McDic:
## Code review round 1

A reviewer went over the diff. It found one genuine defect, several guard gaps, and a handful of leftovers. Everything actionable is fixed in `1cc62c7`; below is what changed and what I deliberately left alone.

---

## Fixed

### Issue 1 (HIGH) — a language switch could restart the WASM equity pass

**The finding:** `HandHistoryCharts` ran the equity pass and built the Plotly figures in *one* effect. Figures carry language, so `language` had to be a dependency of that effect — which dragged the equity pass along with it. A switch mid-analysis superseded the running compute, but `collectAllInDataAsync` spawns up to 8 Web Workers and **has no abort path**, so the old pool kept running while the replacement — seeing a still-empty `equityCache` — could start a second pool for the same hands. Repeated toggling multiplies that.

**How it is fixed:** not by guarding the symptom. The cause was structural, so the layers are now actually separate:

```
data   — the WASM equity pass. Depends on the hand histories and nothing else.
charts — the Plotly figures.   Depends on that data plus the active language.
```

A language switch re-runs only the chart layer. It *cannot* restart the equity pass, because the equity pass has no language dependency — the guarantee is structural, not a matter of timing.

Worth recording honestly: I **could not reproduce the double-pool empirically**, even with 302 all-in hands and a switch fired mid-pass. With WASM plus the preflop cache, the equity pass finishes before the replacement run gets through chip histories and hand usage, so the window never opened. That makes the old code *accidentally* safe, not safe. The split removes the accident. Verified in a browser: switching mid-pass spawns **0 extra workers** (8 total, unchanged) and recomputes no equity, while every figure rebuilds in Korean.

My earlier "bank the cache before the staleness check" change is kept — it is independently correct (equity is keyed by hand id and run-independent, and a second upload landing mid-pass really does supersede a run) — but it is no longer load-bearing, and its comment no longer over-claims.

### Issue 2 (MEDIUM) — nothing verified that `t()` call sites pass the right placeholder names

The types check that the *key* exists. `locales.test.ts` checks that `en` and `ko` agree on the placeholders *inside* it. Neither reads the call site, so

```ts
t('chart.bankroll.tick.buyIns', { initialCapital: r.initialCapital })   // key declares {{capital}}
```

type-checks, tests green, and ships an x-axis whose every tick reads the literal `{{capital}} buy-ins`.

New `callSites.test.ts` scans the source for keys named as literals — **including the ones travelling as `messageKey` / `messageParams`**, which is where most of the interpolation actually lives, so a scanner that only understood `t(` would have missed the important ones — and fails when the values do not match. Confirmed by injecting exactly the bug above: `tsc` still exits 0, and the new test fails.

### Also fixed

- `identityT` folds values into its output instead of silently dropping them, so the same class of error is visible to chart tests.
- `locales.test.ts` now pins **markup parity** too — dropping `<extra></extra>` from a hovertemplate silently breaks the tooltip, and it is the easiest thing in the file to mistake for text.
- Tests pin the language instead of inheriting jsdom's `navigator` locale.
- Deleted `analyzeBankroll` — dead in the web app, and it had grown a `t` parameter purely to satisfy the compiler.
- Stopped shadowing the translator `t` with the tournament loop variable in `rrByRank` / `rreHeatmap` / `historicalPerformance` (`prizePies` was already de-shadowed).
- Dropped the unused `Language` export; fixed a comment in `htmlExport` that had lost its own escape sequence.

---

## Not fixed, with reasons

**The error list is English.** Parser and worker errors (`${file.name}: ${result.error}`) stay untranslated, and `errors.worker` renders eagerly so it freezes the language at error time. This is a deliberate call, not an oversight: those strings are diagnostics that carry file names and raw parsed values, and they are what a user pastes into a bug report. Translating them makes the report harder for me to read without making the failure any clearer to the user. Happy to revisit if you disagree — it is the one place I am least sure.

**The export gate can still fire on a language switch.** Rebuilding figures sets `isComputing`, so hitting Export during the (now brief, WASM-free) rebuild shows the "charts still calculating" modal, and the modal's wording — "may be **missing** from the export" — is the wrong diagnosis for a relabel. It is honest that a rebuild is in progress, the window is short, and fixing the copy properly means teaching the modal to distinguish "computing" from "relabelling". Not worth the scope here.

**Two pre-existing `rrByRank.ts` bugs.** Unrelated to i18n, so left for a follow-up rather than smuggled into this PR:
- the `Rough ITM Cut (~12.5%)` annotation passes a raw `x: 1/8` on a **log** axis, where Plotly wants log10 — so the label lands at ~133 % percentile, off the edge, detached from the line it names. The break-even annotation right below it correctly uses `Math.log10(1)`.
- an all-losses upload leaves `minPercentile === Infinity`, so the axis gets `range: [0, Infinity]` and renders a blank plot instead of the "no data" state the other builders show.

**Korean wording nits.** The reviewer flagged `Marginal → 누적` (*cumulative*, where the term is 주변) and the mixed `바인` / `바이인` spellings. Both come **verbatim from the original Python `translation_values.json`**, i.e. they are the author's own choices, so I have not unilaterally changed them. @McDic — your call.

**jsdom for every test file.** ~4.6 s of environment setup across the suite. Confining it with `environmentMatchGlobs` would save a couple of seconds at the cost of a config knob and a foot-gun for the next test that quietly needs a DOM. Not worth it.

---

CI is green: typecheck, 69 tests, build, pre-commit.

McDic:
## Code review round 2

Scoped to `1cc62c7` — the data/chart split, which no reviewer had seen when round 1 ran. It found **a worse bug than the one the split was fixing, introduced by the split**. Fixed in `7c00103`.

---

## Fixed

### Issue 1 (HIGH) — the export could silently ship without the All-In Equity chart

`isComputing` was two booleans, one per effect. At the moment the equity pass lands, it publishes the data and clears its own flag **in the same commit** — but the effect that turns that data into a figure has not run yet. So for one committed render *neither* flag is set and no equity chart exists.

`App.handleExport` reads `isComputing()` synchronously to decide whether to warn. An export fired in that window produced a hand-history HTML file with the **all-in equity chart missing entirely** — no modal, no warning. That modal exists to prevent precisely this. The reviewer captured the offending render:

```
RENDER calc=false draw=false equity=true      <-- isComputing === FALSE, no equity chart
CHART effect fires                            <-- only now does the flag go back up
```

**Why it happened:** the old single-effect code got the invariant *"there is work outstanding"* for free. Rebuilding it as two hand-maintained flags reintroduced, at the seam between the layers, exactly the class of bug the split was meant to remove.

**The fix is to stop tracking it.** Each figure now carries the data and language it was built from, and "busy" is *derived*: some figure does not match the inputs it should have been built from. That cannot dip — the instant `equity` changes, the equity figure is stale **by definition, in the same commit**.

### Issue 2 (MEDIUM) — chip histories and hand-usage heatmaps were rebuilt on every equity completion

Neither depends on `equity`, but both lived in an effect that had it as a dependency. Measured: `chip=2 usage=2` per upload, and confirmed in the browser — the draw started, got cancelled when equity landed, and redid the same work. `getHandUsageHeatmapsData` is the expensive one. They are now their own effect, keyed on what they actually depend on, and build **once**.

### Issue 3 (MEDIUM) — the progress bar ran backwards

Two concurrent figure layers sharing one percentage: the fast one claimed 100%, then the slow one reported 15%. The bar is now derived too, so it is monotonic.

### Issue 4 (MEDIUM) — the rewritten concurrency had no tests

Fair hit. The harness already existed in `TournamentCharts.test.tsx` (gated `yieldToBrowser`, mocked viz). New `HandHistoryCharts.test.tsx` pins:

- a language switch rebuilds all three figures and **does not recompute equity** — fails against the pre-split version, which did;
- each hand-derived figure is built **exactly once** per upload — fails against `1cc62c7`, which built them twice.

Both verified to fail on the code they guard against, so they are real regression tests rather than decoration.

**One thing I did not test, and said so in the file rather than implying cover:** the one-frame flag dip itself. `act()` flushes React's passive effects synchronously, so no test in this harness can observe a single commit — I tried a MutationObserver commit-sampler and it could not see it either. That window is closed *by construction* (derived state), not by assertion.

### Also fixed

- **Finished the `analyzeBankroll` deletion.** `collectRelativeReturns`, `runBankrollSimulation` and `BankrollAnalysisOptions` had no other caller, and the last was still exported as public surface. `bankrollAnalysis.ts` halves.
- **`callSites.test.ts` now reports a call site it cannot parse instead of skipping it** — skipping would silently stop checking that key, which is the exact vacuity the file exists to prevent.
- **`hasOwnProperty` instead of `in`** — a quoted `'toString'` anywhere in `src/` would have been taken for a key and crashed the suite.
- **The data effect no longer invalidates itself from cleanup.** A re-run that then early-returns would have killed the pass in flight with nothing to take over, wedging `isCalculating` true and the export behind the modal permanently.
- Stopped shadowing `t` in `chipHistories.ts` (round 1 only did the tournament builders); awaited the language pin in test setup.

---

## Not fixed, with reasons

**A second upload landing mid-pass still spawns a duplicate worker pool.** Real, and the reviewer is right that it is the last instance of the class. It is **pre-existing** — `equityCache` is only written after the whole batch resolves, so any pass starting while another is in flight sees an empty cache. `1cc62c7` fixed the *language* trigger structurally; the *upload* trigger predates this branch and needs a different fix (an in-flight registry beside the cache, or banking incrementally). That is a change to the equity module's contract, not to i18n, and I would rather not smuggle it in here. Happy to do it as a follow-up.

**`progress.chart.error` is never actually shown to the user.** Both catch blocks set the message, but the progress block is gated on `isComputing` — so setting the error is exactly what unmounts it. Errors reach `console.error` only. Pre-existing shape; fixing it properly means a distinct error surface rather than reusing the progress bar. Worth a follow-up, not worth widening this PR.

**Korean wording** (`Marginal → 누적`, mixed `바인`/`바이인`) — still verbatim from your original Python `translation_values.json`. @McDic, your call.

---

CI green: typecheck, **72 tests / 11 files**, build, pre-commit. Lint unchanged at 7 pre-existing errors.

PR #31, opened 2026-07-13
Fix two Plotly coordinate bugs in the RR-by-Rank chart

McDic:
Two pre-existing bugs in `rrByRank.ts`, found while reviewing the i18n PR (#30) and deliberately kept out of it.

## 1. The "Rough ITM Cut (~12.5%)" label never rendered

On a log axis, Plotly reads a **shape's** coordinates as data but an **annotation's** as **log10**. That split is counterintuitive enough that I verified it against a rendered plot rather than trusting the docs — a shape at `0.125` lands exactly on the `0.125` data point, while an annotation only does so at `log10(0.125)`:

| | raw `0.125` | `log10(0.125)` |
| --- | --- | --- |
| **Shape** | **px 311 ✓** (on the data point) | px 7858 ✗ |
| **Annotation** | *never rendered* (clipped) | **px 283 ✓** |

The label was passed `1 / 8` raw, which put it at `10^0.125` ≈ **133%** — outside the axis range, so it was clipped and never drew. Its line is a shape and was correct as-is; it is left alone.

Confirmed in a browser against real data: **before**, the label is absent from the DOM entirely; **after**, it renders inside the plot area, **2 px** from the line it names.

The break-even annotation directly below it already used `Math.log10(1)`, which is what made the inconsistency visible in the first place.

## 2. A player who has never cashed got a blank plot with a nonsense axis

Every axis bound is derived from tournaments that paid, and the filter above keeps only those. With none of them, `minPercentile` kept its `Infinity` sentinel:

```
xaxis.range = [0, Infinity]      // Plotly discards this, falls back to a default axis
            = [0, null]          // in the HTML export — JSON.stringify(Infinity) is null
```

So instead of the "no data" state the bankroll and all-in-equity charts already show, you got an empty plot with a meaningless 100%–1000% percentile axis. It now shows a no-data state (new key `chart.rrByRank.noData`, EN + KO).

## Tests

New `tournament.test.ts` pins both, and both were verified to **fail** against the unfixed code (`expected 0.125 to be close to -0.903…`; `expected [ 'xaxis.range[1] = Infinity' ] to deeply equal []`).

It also adds a shared guard that **every** tournament chart hands Plotly only finite numbers on degenerate data (no tournaments, nothing cashed, a single tournament). The non-finite-axis class is not specific to this chart, and a thin or unlucky history is a real user. The other four charts pass it today.

## Verification

`npm run build`, `npm test` — 85 tests / 12 files. Driven end-to-end in a browser with 120 tournament summaries; lint unchanged at 7 pre-existing errors.

### DIFF
--- a/web/src/i18n/locales/en.json
+++ b/web/src/i18n/locales/en.json
@@ -124,6 +124,7 @@
 
   "chart.rrByRank.name": "RR by Rank",
   "chart.rrByRank.title": "RR by Rank Percentile",
+  "chart.rrByRank.noData": "No cashes yet — nothing to plot",
   "chart.rrByRank.legend.rrByPercentile": "RR by Percentile",
   "chart.rrByRank.legend.perr": "PERR (Percentile × RR)",
   "chart.rrByRank.legend.trendline": "RR Trendline (top 12.5%)",
--- a/web/src/i18n/locales/en.json
+++ b/web/src/i18n/locales/en.json
@@ -67,6 +67,7 @@
 
   "chart.historical.name": "Historical Performance",
   "chart.historical.title": "Historical Performance",
+  "chart.historical.noData": "No tournaments to plot",
   "chart.historical.legendGroup.profitsRakes": "Profits & Rakes",
   "chart.historical.legendGroup.profitableRatio": "Profitable Ratio",
   "chart.historical.legendGroup.avgBuyIn": "Avg Buy-In",
--- a/web/src/i18n/locales/ko.json
+++ b/web/src/i18n/locales/ko.json
@@ -124,6 +124,7 @@
 
   "chart.rrByRank.name": "순위 백분위별 RR",
   "chart.rrByRank.title": "랭킹 백분위수별 RR",
+  "chart.rrByRank.noData": "아직 상금을 획득한 토너먼트가 없습니다",
   "chart.rrByRank.legend.rrByPercentile": "백분위별 RR",
   "chart.rrByRank.legend.perr": "PERR (백분위 × RR)",
   "chart.rrByRank.legend.trendline": "RR 추세선 (상위 12.5%)",
--- a/web/src/i18n/locales/ko.json
+++ b/web/src/i18n/locales/ko.json
@@ -67,6 +67,7 @@
 
   "chart.historical.name": "과거 성적",
   "chart.historical.title": "과거 성적",
+  "chart.historical.noData": "표시할 토너먼트가 없습니다",
   "chart.historical.legendGroup.profitsRakes": "수익 & 레이크",
   "chart.historical.legendGroup.profitableRatio": "멘징 이상 벌은 비율",
   "chart.historical.legendGroup.avgBuyIn": "평균 바이인 금액",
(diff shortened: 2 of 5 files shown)


McDic:
## Code review round 1

Reviewer independently reproduced both bugs and the fix — rendering Plotly 3.3.1 in headless Chromium and probing pixel positions, then diffing the real branch figure against the real master figure. It confirmed the log10-vs-data asymmetry precisely, and audited **every** shape and annotation in **every** chart against whether its axis is `type: 'log'`: exactly one was wrong, which is what this PR fixes. **Approved**, with findings below all addressed in `36ca3ec`.

---

## Fixed

### Issue 1 (MEDIUM) — the "every chart" guard omitted the one case that fails it

Fair hit, and the most useful finding here. The block titled *"every tournament chart, on degenerate data"* was calibrated to the cases that happened to pass. It left out:

```
getHistoricalPerformanceData([])   →   yaxis3.range = [Infinity, -0.95]
```

`Math.min(...[])` is `Infinity`, `Math.max(...[])` is `-Infinity`, and the shapes anchor to `netProfit[-1]`, which is `undefined`. Unreachable from the app (`TournamentCharts` returns early on an empty set) — but a chart builder should be safe on its own inputs, and **a guard whose promise is bigger than its coverage is worse than no guard.** The case is now in the list, and the builder handles it the way `bankrollAnalysis` and `allInEquity` already do.

### Issue 2 (LOW) — `data.length === 0` was the symptom, not the invariant

Also right, and it made the fix better. `rankPercentile` is `myRank / totalPlayers`, and neither is sanity-checked — so a zero-player field still put `Infinity` into the x-axis range **and into the plotted coordinates and `customdata`**, through a different door:

```
BAD  totalPlayers=0        xaxis.range[1] = Infinity
BAD  myRank=0              xaxis.range[1] = NaN
BAD  one good, one corrupt traces[0].x[0] = Infinity
```

Tightening the existing filter to require a finite, positive percentile **subsumes the empty case in one clause** and stops the coordinates too. Both new cases were added because they failed.

### Issue 3 (LOW) — the guard checks layouts only, and didn't say so

It read as an oversight. It was a decision, and now says which: **NaN is legal in trace data**, where Plotly reads it as a gap — `getRREHeatmapData` emits it deliberately for a zero RRE. Layouts are the place non-finite numbers are fatal.

But the reviewer's point exposed a real gap underneath: `Infinity` in a trace's **x** is *not* a gap marker, it's a coordinate Plotly cannot place, and the export writes it as `null`. So there is now a second assertion for exactly that, and it is what caught the zero-player case above.

### Nits

- The ITM shape has a `name`, so the test can say **which** shape it means rather than describe it as "the one that isn't at x=1".
- The happy-path bankroll case is out of the block titled "degenerate data".

---

## Not fixed, with a reason

**The no-data plot draws Plotly's default −1…6 grid behind the message.** Cosmetic, and it is exactly what `bankrollAnalysis` and `allInEquity` already do — so fixing it here alone would make the three inconsistent. Worth doing to all three at once, in a PR about that; not worth a one-off divergence in a PR about coordinate bugs.

---

CI green: typecheck, **102 tests / 12 files**, build, pre-commit. Lint unchanged at 7 pre-existing errors.

PR #32, opened 2026-07-13
Compute each hand's equity once, and show the user when a chart fails

McDic:
The two follow-ups agreed on #30. Both were raised in review there and deliberately kept out of it.

---

## 1. Each hand's equity is computed once, ever

All-in equity is the one genuinely expensive thing the app does: `collectAllInDataAsync` spawns up to **8 Web Workers**, each loading a 686 KB preflop cache and initialising WASM, and **none of them can be aborted** once started.

The cache guarding that was only written once a whole batch resolved. So a second upload landing mid-pass saw every in-flight hand as uncached and could start a **second pool** on hands the first pool was already grinding — two unabortable pools fighting over the CPU to do the same WASM twice.

New `equityStore.ts` owns the cache *and* a record of what is being computed right now, claimed **synchronously, before anything is awaited**. A pass waits for hands another pass already holds rather than recomputing them; only genuinely new hands reach a worker.

Its unit tests demonstrate the race directly. With the in-flight tracking removed, loading `[a, b, c]` while `[a, b]` is still in flight hands the whole lot to a fresh pass:

```
expected [ ['a','b'], ['a','b','c'] ]   // recomputes a and b
to equal [ ['a','b'], ['c']       ]   // only c is new
```

### Being straight about the evidence

**I could not reproduce the duplicate pool in a browser.** Parsing a second upload takes longer than the equity pass takes to finish, so the window never opened — the same *accidentally safe, not safe* shape as the language-switch trigger fixed in #30. The store makes the invariant hold **by construction** rather than by timing.

Measured after the change, with a second upload dropped mid-pass (160 files, 302 all-in hands):

```
equity workers spawned : 16   (8 for batch 1, 8 for batch 2's own new hands)
hands sent to workers  : 302
distinct hands         : 302
recomputed twice       : 0     <-- no hand computed more than once
```

## 2. A failed chart is no longer silent

Both chart components caught their failures, cleared `isComputing`, and wrote `progress.chart.error` into the progress state — but the progress block **only renders while work is outstanding**. So writing the failure there was precisely what unmounted the thing that would have displayed it.

Four catch blocks, all dead ends. The only trace a user got of a broken chart was a `console.error` they never opened.

Failures now have their own state and render as a `role="alert"` banner. Verified by injecting a WASM failure:

```
banner (en): All-in equity could not be calculated. See the browser console for details.
banner (ko): 올인 에퀴티를 계산하지 못했습니다. 자세한 내용은 브라우저 콘솔을 확인해주세요.
role       : alert
```

...while the other charts carry on rendering.

## Also

A test that **no key in `en.json` is unreferenced**. Rewiring the equity path orphaned two keys (`progress.chart.error`, `progress.chart.equityCached`) and I only caught them by hand. Translators should not spend effort on text that nothing displays, and the type system only checks the other direction — that a key *used* exists.

## Verification

`npm run build`, `npm test` — 77 tests / 12 files. Every new test verified to **fail** against the unfixed code. Driven in a browser for both the dedup measurement and the injected failure. Lint unchanged at 7 pre-existing errors.

### DIFF
--- a/web/src/App.css
+++ b/web/src/App.css
@@ -352,6 +352,18 @@
   color: #666;
 }
 
+/* Chart Failure State — shown independently of the progress bar, which only exists
+   while work is still outstanding. */
+.chart-error {
+  padding: 1rem 1.25rem;
+  margin-bottom: 1.5rem;
+  background: #2a1a1a;
+  border: 1px solid #5a2a2a;
+  border-radius: 12px;
+  color: #f2a0a0;
+  text-align: center;
+}
+
 /* Chart Loading State */
 .chart-loading {
   padding: 2rem;
--- a/web/src/components/HandHistoryCharts.test.tsx
+++ b/web/src/components/HandHistoryCharts.test.tsx
@@ -55,8 +55,11 @@ const equity = vi.hoisted(() => {
     inFlight: () => release !== null,
   }
 })
-vi.mock('../visualization/handHistory/allInEquityAsync', () => equity)
-vi.mock('../visualization/handHistory/equityStore', () => equity)
+vi.mock('../visualization/handHistory/equityStore', () => ({ loadEquity: equity.loadEquity }))
+vi.mock('../visualization/handHistory/allInEquityAsync', () => ({
+  calculateLuckScore: equity.calculateLuckScore,
+  createAllInEquityChart: equity.createAllInEquityChart,
+}))
 
 const gates = vi.hoisted(() => {
   let waiting: Array<() => void> = []
@@ -75,9 +78,8 @@ let container: HTMLDivElement
 let root: Root
 let ref: RefObject<HandHistoryChartsRef | null>
 
-// The equity store is module-level and survives between tests, so every test needs hand
-// ids of its own — otherwise the second test finds the first test's equity already there.
 let idSeq = 0
+/** Fresh ids per test, so nothing can accidentally depend on another test's hands. */
 function makeHands(count: number): HandHistory[] {
   const batch = ++idSeq
   return Array.from({ length: count }, (_, i) => ({ id: `t${batch}-h${i}` }) as HandHistory)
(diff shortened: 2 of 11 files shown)


McDic:
## Code review round 1

Reviewer said **request changes** on 1 HIGH + 3 MEDIUM, and it was right about all of them. Fixed in `612d912`.

It also strengthened the case for the PR: it found **two reachable routes** to the race I said I couldn't reproduce — the file input is disabled on `isLoading` (the *parse* worker's flag), not during the equity pass, so a large first file plus a small second one lands upload 2 mid-pass; and `ErrorBoundary resetKeys={[handHistories]}` remounts the component mid-pass. My non-repro was a fact about my test data's timing, not about reachability.

---

## Fixed

### Issue 1 (HIGH) — a dead worker poisoned the whole session. My regression.

`collectAllInDataAsync` resolves a worker's promise on a `message` and rejects on an `error`. A worker **killed outright** — the browser reclaiming one of eight, each holding a 686 KB cache and a WASM instance, which is the very cost this PR exists to avoid paying twice — fires **neither**. Its promise never settles, and neither does the `Promise.all` over the pool: no result, no rejection, nothing to catch.

That was already true before this PR. What the PR did was make it **permanent**: dedup has every later caller *join* the dead pass, where before, a re-upload spawned a fresh pool and self-healed. Reviewer's repro:

```
upload1: pool for [allin_a]   <- worker dies silently
upload2: pool for [allin_b]   <- healthy, would have succeeded
upload3: pool for [allin_c]   <- healthy
resolved loads: (none — every caller hangs forever)
inFlight still claims: [allin_a, allin_b, allin_c]
```

Workers now have a **watchdog**. It bounds *silence*, not runtime — the timer restarts on every message, so a genuinely long batch is never mistaken for a dead one. A presumed-dead worker is terminated and its pass rejected, which is what lets the store release its claim so a retry can re-spawn. Two new tests prove it: **with the watchdog removed, both hang until vitest kills them at 5 s.**

### Issue 2 (MEDIUM) — the progress bar froze and lied, in exactly the path the store exists to fix

A caller that joined someone else's pass reported only *its own* numbers — so the bar sat at a **completed** count (`1/1`) while the pass holding the hands it needed ground on invisibly. Passes now carry live progress, and a caller reports the **union across every pass it is waiting on**. The store's test double now drives `onProgress`, without which this was invisible to the suite.

### Issue 3 (MEDIUM) — the error banner never cleared

Only the data effect cleared it, and its deps don't include `language` — so a chunk that 404s once would leave "Some charts could not be generated" on screen **forever**, even after a language switch rebuilt all three charts perfectly.

The reviewer also called the naive fix wrong, and it is: one slot for three independently-retrying layers means a recovering layer would wipe another layer's still-live failure. Failures are now tracked **per layer**. New test, verified to fail without it.

### Issue 4 (MEDIUM) — an error banner sitting above a live progress bar, forever

A failed figure layer left `handFiguresStale` true, so `isComputing` stayed true: a red banner above a bar reading "Generating hand usage heatmaps…" at 90%, permanently, with the export gate held shut on charts that were never coming. **A failed layer is not "outstanding" — it is finished, badly.** It no longer counts, so the bar goes away and the user can export what did build.

### Also

- **Claim the hands strictly before the pass runs.** They were previously first only because nothing in `collectAllInDataAsync`'s synchronous prologue happened to yield — which is true, but an invariant this module rests on should not depend on it. The claim now precedes the pass body unconditionally.
- The store's test double models **eligibility**, so it stops hiding that a hand with no all-in is never cached and is re-offered every call.
- The unused-key scanner **ignores comments** — these comments name keys constantly, and a key mentioned only in prose was counting as used.
- Key ordering: the two failure keys no longer split the `noTournamentData` / `noHandHistoryData` pair.

---

## Not fixed

**`cache` grows unboundedly per session.** Pre-existing, and bounded in practice by how many hands a person uploads in one sitting.

---

CI green: typecheck, **82 tests / 13 files**, build, pre-commit. Lint unchanged at 7 pre-existing errors.

PR #33, opened 2026-07-13
Preflop situation ledger: what each decision paid, against folding

McDic:
Phase 1 of the situation charts we discussed. Adds a new **Preflop Situations** tab with the *situation ledger*: for every preflop decision — RFI, open-jam, open-limp, over-limp, iso-raise, blind defence, 3-bet, squeeze, call-vs-3-bet, 4-bet, call-vs-4-bet, 5-bet — what it paid, split by position.

No parser or WASM change. `actionsPreflop` already retains the complete ordered action sequence per player with sizes and all-in flags, so this is pure aggregation plus a figure.

> **Note:** an earlier revision of this description quoted a figure from a row the chart could not actually draw. See the review iteration comment below for what went wrong and how it was fixed. Everything here reflects the current state.

## The metric: Δbb, not raw bb

Folding is not worth zero — it forfeits what you already posted. Measured as raw profit, **every blind-defence bucket comes out negative** for reasons that have nothing to do with how well the hand was played, and no honest comparison against an open-raise bucket is possible. So each decision is scored against folding *right there*:

```
Δ = net result + chips already committed
  = collected − chips voluntarily invested from this node on
```

The antes and blinds **cancel exactly**, so the metric needs no knowledge of the ante structure (uniform, as online tournaments actually use, or big-blind) — only Hero's own posts.

Two consequences:

- **Folding scores exactly 0, at every node.** An identity, not an expectation. It is the load-bearing test — verified across **58,582 real fold decisions** in a 72,840-hand corpus, all exactly zero. If the chip accounting is ever off by an ante, it breaks there rather than in a chart.
- **Within a bucket, Δbb is raw bb shifted by a constant.** It is not a new derived quantity that could quietly be wrong; it moves the zero, which is the whole point.

## Colour carries significance, not sign

A row is blue or red only when its 95% interval clears zero; otherwise grey. On a thin sample, 3-betting scored **+4.79 ± 5.33bb** — paint that blue by its sign and the chart reports a large edge the data does not support. Grey says the only true thing: this cannot be told apart from folding.

Colour is never the sole cue — the interval visibly crosses the zero line, the legend names all three states, and `n` is printed into every row label. Palette validated against the light surface Plotly paints on (worst adjacent CVD separation ΔE 74.6, all steps ≥ 3:1); blue↔red rather than the green↔red poker convention, which is exactly the pair deuteranopes cannot separate.

## Rows compare within a group, not down the chart

Real data surfaced a trap worth naming. Δ is measured against folding, and what folding costs depends on how much is already in front of you. "Open raise · BTN" scores around **+1bb** (folding costs nothing). "Call vs. 4-bet · BTN" scores **+31bb** — not a talent for calling 4-bets, but the geometry of a spot where folding surrenders a 3-bet into a large pot. Families are therefore ordered by pot size and the caption says so explicitly, because a reader who ranks every row against every other will conclude they should call more 4-bets, and the data does not say that.

## Guardrails

The failure this feature most needs to avoid is a chart that is *quietly* wrong, so two invariants are asserted rather than assumed:

- **Fold-scores-zero**, at every decision node (`preflopSituation.test.ts`).
- **Family coverage**: every `(context, action)` the classifier can emit is matched by *exactly one* row family, or is a fold (`situationLedger.test.ts`). Without this, a decision is computed, correct, and silently dropped — no row, and not counted among the hidden ones either, so the chart looks complete. That is exactly what the first revision did to over-limps and every 4-bet-facing decision, and it hid a real losing spot.

Filters (opener position / starting stack / players dealt in / minimum sample) all narrow the same pre-classified array, so a dropdown change never re-walks the hand histories.

## Also

- Position labels hoisted from `chart.handUsage.position.*` to a shared `position.*` namespace, now that two charts use them.
- English-only strings; `ko.json` falls back per the existing i18n contract (`locales.test.ts` treats a missing key as a legitimate partial translation).
- Export path now picks the ref by tab rather than by "not tournament", which would otherwise have exported the wrong tab's charts.

Phase 2 — the situation-filtered 13×13 hand grid — follows in a separate PR.

## Verification

- **163 tests pass**, including the i18n guards (`callSites.test.ts` no-dead-keys, `locales.test.ts` fallback).
- `npm run build` green; new files lint clean.
- Figure rendered against the real 72,840-hand corpus and inspected — caught three layout bugs no test would have (row labels stamped across the bar panel, caption overflow, legend/axis-title collision), all fixed.

## Known tradeoff

Both panels share one x-axis, so the wide intervals on the low-sample 4-bet rows compress the RFI rows near zero. Honest, but not maximally readable — raising the minimum-sample filter drops the noisy rows. Per-group axes (small multiples) are the fix if it bothers you.


McDic:
## Code review iteration 1

Reviewer found **2 HIGH, 4 MEDIUM, 7 LOW**. Fixed in 5eea73d. The headline finding was real and serious, so leading with it.

---

### Issue 1 (HIGH) — Whole classes of decisions were computed and silently dropped ✅ FIXED

`FAMILIES` did not cover what `classifyHand` emits. **Over-limps** (`limped/call`) and **every 4-bet-facing decision** (`fourBetPlus/*`) were classified, scored correctly, and then thrown away — they became no row, *and* were not counted among the `hidden` ones, so the chart looked complete.

This is the worst possible failure mode for this feature, because the symptom is an **absence**. Confirmed against the real 72,840-hand corpus: it was hiding an **over-limp from the cutoff running at −1.48bb** — precisely the sort of leak the chart exists to surface.

It gets worse: the module doc *and my PR description above* both quoted `+18.17bb` / `+29.77bb` for calling a 4-bet. **Those rows could not render.** I read those numbers off a dump of the classifier, never off the chart, and nothing connected the two.

Fixed by adding the missing families (over-limp, call vs. 4-bet, 5-bet+) and renaming the raise+caller call family — it also catches a BB defending multiway, which is not a cold call.

Verified on real data after the fix: **orphaned non-fold decisions: 0**, and over-limp / call-vs-4-bet now produce rows. (5-bets exist but at n=62 across 8 positions they legitimately fall below the sample threshold and land in `hidden` — correct behaviour.)

### Issue 2 (HIGH) — `situationLedger.ts` had no tests ✅ FIXED

Which is *why* Issue 1 shipped. Added `situationLedger.test.ts`, and the load-bearing one is the coverage test:

> every `(context, action)` the classifier can emit is matched by **exactly one** family, or is a fold.

That is the only thing that can catch a silent drop. Also covers `summarize`, `colorOf`, the `hidden` counter, family ordering, and the heads-up separation. 138 → 163 tests.

### Issue 3 (MEDIUM) — A preflop `bet` would have broken the fold identity ✅ FIXED

`'bet'` was in `VOLUNTARY` (so it created a situation) but had a **no-op case** in the commitment switch — while `getHandHistoryTotalChipsPut` *does* add it. Hero's tracked commitment would desync for the rest of the hand, putting a nonzero Δ on every later node **including folds** — silently breaking the one identity this whole module rests on.

Unreachable today (GG never writes a preflop `bet`), but as the reviewer put it, it did the worst of both readings. Now the hand is **dropped** with a comment explaining why: `bet` carries an increment where `raise` carries a *to* total, so there is no coherent reading. An empty chart is a visible failure; a subtly wrong one is not.

### Issue 4 (MEDIUM) — Heads-up pooled into the short-handed rows ✅ FIXED

`getHandHistoryOffsetFromButton` reports the HU button as SB (the button *is* the SB heads-up), so an HU steal — where opening most of the deck is correct — was averaged into the same **"Open raise · SB"** row as a 6-max SB open. Late-MTT play is heavily short-handed, so not a corner case, and the caption explicitly tells you to compare across positions within a family — so this was the live "silently wrong chart" risk.

`PreflopSituation` now carries `tableSize`, and there is a **Players dealt in** filter (Heads-up / 3–6 / 7+). Covered by a test that shows the pooled mean is misleading and the split means are right.

### Issue 5 (MEDIUM) — `colorOf` painted a certain row grey ✅ FIXED

Keyed on `ci95 === 0` as a proxy for `n < 2`, but a zero-variance sample also has a zero-width interval — so the one kind of row we are *certain* about got painted "indistinguishable from folding". Keys on `n` now.

### Issue 6 (MEDIUM) — ~90 array sweeps per dropdown change ✅ FIXED

`buildLedgerRows` did 1 + 11 + up to 88 filter passes with intermediate allocations, synchronously inside a `useMemo`, on a six-figure array. Replaced with one bucketing pass. Rows are still emitted in family order (stable across datasets, and the depth ordering the caption promises actually holds).

### LOW ✅ FIXED

Dropped the unread `handId` / `openerOffset` fields; unexported `clampOffset`; reused the existing `.no-data` class instead of a duplicate `.chart-empty`; removed the redundant `activeTab === 'tournament'` re-test in `App.tsx`.

---

## Not fixed, with reasons

**`cards` and `raiseToBB` are unread (LOW).** Kept deliberately. `cards` is what the phase-2 hand grid keys on, and `raiseToBB` answers the size question that motivated this work ("called a ≤3bb open?"). Both land in the next PR. I'd rather not churn the type twice — but if you'd prefer them removed and re-added, say so and I'll drop them.

**`ko.json` regresses to ~82% coverage (LOW).** Intentional, and what we agreed: English-first, with `fallbackLng: 'en'`. `locales.test.ts:107-119` explicitly treats a missing key as a legitimate partial translation. Korean strings can land any time without touching TypeScript.

**Left margin may clip a long translation (LOW).** Real but not yet triggered — every row label fits in English. Worth revisiting when Korean lands, since the fix depends on the actual string lengths.

**7 pre-existing eslint errors** in `htmlExport.ts`, `parser/utils.ts`, `allInEquityAsync.ts`, plus one `setState`-in-effect error in `App.tsx`'s auto-switch-tab effect. All present on `master`, none in code this PR touches. Out of scope.

---

## State

- **163 tests pass** (was 138), `npm run build` green, new files lint clean.
- Re-verified on the real 72,840-hand corpus: fold identity holds across **58,582 fold decisions, all exactly zero**; **0 orphaned decisions**.

McDic:
## Code review iteration 2

Reviewed the delta since round 1 (`20354af..HEAD`). **No CRITICAL, no HIGH.** The riskiest item — the `generateExportHTML` signature change — was verified fully migrated and non-regressive for both pre-existing export paths, with the `</script>` breakout defence intact and now extended to captions. The `buildLedgerRows` single-pass rewrite was confirmed behaviour-identical to the nested-filter version it replaced.

What round 2 actually found was **three places where the code was right and quietly withheld something from the reader**. All fixed in 21ae6dc.

---

### Issue 1 (MEDIUM) — The dropped hand was invisible, and untested ✅ FIXED

A preflop `bet` makes the walk drop the hand, and my comment justified it with *"an empty chart is a visible failure; a subtly wrong one is not."* The reviewer correctly pointed out **that is false**: it drops a **hand**, not the chart. If GG ever emitted a preflop `bets` line for some subset of hands, those hands would vanish with no counter, no error, and no empty chart — a complete-looking ledger built on a silently truncated corpus, which is the exact outcome the bail-out claims to prevent.

- `classifyHandHistories` now returns `{ situations, droppedHands }`, and the caption reports the count with an explicit "this should be zero" note.
- Added the tests the bail-out never had. A behaviour that discards data and has zero coverage is one refactor away from discarding more. Also pinned that a **walk** (no decision, but nothing refused) does *not* count — otherwise the warning would cry wolf on every ordinary session.
- Verified on the real 72,840-hand corpus: **droppedHands = 0**, as expected.

### Issue 2 (MEDIUM) — The default view still did the thing `tableSize` was added to prevent ✅ FIXED

Sharp catch. The `TableBucket` doc calls HU-pooled-into-SB a bug; the test that proves it literally names the pooled number *"the misleading average"* — and `DEFAULT_FILTERS.tableBucket` is `'any'`, so **that is exactly what the chart shows** until the reader touches the filter.

Pooling everything is a legitimate "show me everything" default. *Not saying so* is not. The caption now states that positions pool table sizes, that heads-up puts the button on the small blind, and which filter splits them.

### Issue 3 (LOW, but the best catch of the round) — The export flattened the one line that carries the meaning ✅ FIXED

`patchAxesForLightTheme` spread its colours *over* the chart's own, so the ledger's deliberately heavy `#898781` zero line — the line the caption tells you to read the whole chart against — came out of the exporter as a `#bbb` hairline no darker than a gridline. **In the exported file only.**

They are **defaults** now, not overrides (`{...axisDefaults, ...src[key]}`): a figure that styled an axis on purpose keeps it, and everything else still gets its light-background default. Pinned with a test.

### Also fixed (LOW)

- Stale comment still describing the annotation-era caption stacking, in a file where comments are the spec.
- `MIN_SAMPLE_CHOICES` retyped in the component instead of living beside the filter it belongs to (and duplicating the `30`).
- React keys drawn from translated caption text — two identical translations would collide.
- `"1 rows are hidden"`. Reworded to sidestep pluralization, which i18next cannot do here anyway without tripping the `callSites.test.ts` no-dead-keys rule.
- A length-check masquerading as an ordering assertion in my own test.
- Added a two-non-empty-section export test — the multi-section render and plot-call `.join()` paths had no coverage, and prefix uniqueness went from structurally impossible to a caller's promise.

---

## Not fixed, with reasons

**`raiseToBB` and `cards` are still unread (LOW, repeat from round 1).** Deliberate. `cards` is what the phase-2 hand grid keys on and `raiseToBB` answers the size question that motivated this work ("did I call a ≤3bb open?"). Both are consumed by the next PR. The memory cost is a number and a reference to an already-existing array per situation — not a copy. I'd rather not churn the type twice, but say the word and I'll strip them.

**7 pre-existing eslint errors** (`htmlExport.ts` `<\/script>` escapes, `parser/utils.ts`, `allInEquityAsync.ts` unused vars, plus a `setState`-in-effect in `App.tsx`'s auto-switch-tab). All on `master`, none introduced here. Out of scope for this PR.

---

## State — ready to merge

- **173 tests pass** (138 → 163 → 173 across the two review rounds), `npm run build` green, new files lint clean.
- Re-verified on the real corpus: fold identity holds across **58,582 fold decisions, all exactly zero**; **0 orphaned decisions**; **0 dropped hands**.

PR #34, opened 2026-07-13
Profit by hand class: break a ledger row into the hands it is made of

McDic:
Phase 2 of the situation charts. Adds a second chart under the ledger: for one chosen action from one chosen position, **which hands actually made money?**

It is not the 13×13 grid we planned, and the reason is the point of this PR.

## The grid does not work at this data volume — measured, not guessed

Before building it I measured the real per-cell density on the 72,840-hand corpus:

| Scope | decisions | median hands/cell | cells reaching n≥30 |
|---|---|---|---|
| Open raise · BTN | 881 | **6** | **0 of 169** |
| 3-bet vs. open · CO | 322 | **4** | 1 of 169 |
| Flat / defend · BB | 1322 | **7** | 3 of 169 |

A 13×13 grid slices a bucket into 169 cells, and Δbb is heavy-tailed — one cracked ace moves a cell by fifty big blinds. **Four hands is not an estimate of anything**, and a diverging colourscale would have rendered that non-information in confident red and blue. That is the "quietly wrong chart" failure this whole feature has been built to avoid, just one level down from the row. It needs roughly 20–30× more hands.

(Frequency *would* survive the 169-cell split — a rate converges far faster than a heavy-tailed mean. That is exactly why the existing VPIP heatmap can be a grid and this cannot.)

## Classes are the same question at a resolution the data can answer

The bucket that gives 7 hands per *cell* gives **109–420 per class**, and the answer becomes legible. Real output, BB-defence pooled across positions:

| Class | n | avg Δbb | 95% CI | |
|---|---|---|---|---|
| Pair 66–99 | 252 | **+4.01** | ±2.79 | beat folding |
| Offsuit connector | 183 | **+2.55** | ±1.68 | beat folding |
| Offsuit broadway | 420 | +1.68 | ±1.90 | inconclusive |
| Suited trash | 117 | **−0.85** | ±2.08 | worst class |

Thirteen classes: pairs by height, then suited and offsuit each split into broadway, ace-x, connector, gapper, trash.

**Ordering is taxonomic, not by profit.** A chart that reshuffles its rows every time you change a filter is one you cannot compare against the one you were just looking at.

## Guardrails

Same shape as the ledger's family-coverage test, because the defect is the same: an absence, which is the one thing a chart cannot show you.

- **`classOfHand` partitions the deck.** All **1326** two-card combinations land in exactly one class, and no class is unreachable — both asserted. A hand matching no class would be computed, correct, and silently dropped from every bucket.
- Boundary cases pinned the way a poker player would read them: `AKs` and `KQs` are broadway, not connectors; `A2s` is a suited ace, not trash; `K2s` *is* trash, because broadway needs both cards.
- Decisions whose hole cards the history never showed are **counted and reported**, not swallowed.

## Refactor

The two charts are the same figure over different rows, so the figure, the statistics, the significance colouring and the filters now live in `deltaFigure.ts` and `situationFilters.ts`. Two charts on one screen that scored significance differently, or coloured it differently, would be a lie told quietly.

## Verification

- **188 tests** pass (173 → 188), `npm run build` green, new files lint clean.
- Rendered against the real corpus and inspected.

## Deferred (agreed)

Korean i18n for the situation tab, and carrying the active filter values into the exported HTML — both to a follow-up PR.


McDic:
## Code review iteration 1

**No CRITICAL, no HIGH.** The riskiest change — rewriting the merged, reviewed `situationLedger.ts` on top of the extracted modules — the reviewer did not eyeball: they restored `master`'s version alongside the new one and ran both over 20,000 randomized situations across **800 filter combinations**, diffing the full `{traces, layout, caption}` JSON. **Identical in every one.** That is the right way to check a refactor and I'm glad it was done that way.

Three MEDIUMs, all fixed in ff47197. The first one is a genuine poker error on my part and it **changed the chart's answer**.

---

### Issue 1 (MEDIUM) — "Suited trash" contained K9s ✅ FIXED

My gap-based rule made K9s four ranks apart and therefore *trash*, in the same row as `72s`. The reviewer enumerated it: `suitedTrash` was **36 hands** — `K9s K8s … K2s Q8s … J7s … T6s …` — so the row's actual **mass was K-x and Q-x**, not garbage. The ace got a class of its own; the king fell off a cliff. Someone reading "trash defences break even" would have started flatting `92s`.

Added **`broadway-x`** (exactly one broadway card, unconnected), tested *after* connectors so `T9s` stays a suited connector rather than becoming a ten-with-a-nine. Gapper and trash now mean what they say: no broadway card at all. Suited trash is `95s`–`72s` — ten hands, every one genuinely trash.

**This changes the finding, which is the point.** On the real corpus, BB defence:

| | class | avg Δbb | n |
|---|---|---|---|
| **before** | Suited trash | −0.85 | 117 ← the blend |
| **after** | Suited broadway-x | **−0.46** | 161 ← the actual weak class |
| | Suited trash | *hidden* | <30 ← barely defended |

The old headline was an artifact of the blend. The real soft spot is **marginal broadway**, not trash. A bad taxonomy did not just blur the chart — it gave the wrong answer.

### Issue 2 (MEDIUM) — Class coverage enforced in only one direction ✅ FIXED

The test asserted every *listed* class is reachable, but never that every *emitted* class is listed. `buildHandClassRows` iterates the list, so a class the classifier emits but the list omits would be bucketed and then never read — vanishing with no `hidden` count, no `noCards` count, and every test still green. Exactly the invisible-absence failure the ledger's family-coverage test exists to prevent, one level down.

- Added the reverse assertion (emitted ⊆ listed).
- Made the label map a **`Record<HandClass, TranslationKey>`** so the *compiler* enforces that a fourteenth class cannot ship without a name. An array of pairs let a missing member through silently, and `KEY_OF.get(c)!` would have rendered it through `t(undefined)`.

### Issue 3 (MEDIUM) — The empty chart threw away the number that explains it ✅ FIXED

`hidden` and `noCards` were computed and pushed into the caption — but the caption only mounted when there were rows, so they were discarded **exactly when the reader needs them**. "You squeezed from UTG six times" and "you have never squeezed from UTG" looked identical, and only one of them is fixed by lowering the threshold.

The caption now renders regardless of whether there are rows, and carries an `inScope` count. The class chart also gets its own empty copy — the ledger's *"widen the filters"* is the wrong advice when the fix is "pool the positions or pick an action you take more often". The duplicate empty message is gone too.

### LOW ✅ FIXED

- Unexported what the extraction made needlessly public (`BEAT_FOLD`/`LOST_TO_FOLD`/`INCONCLUSIVE`/`colorOf`/`DeltaFigureOptions`, `DEFAULT_MIN_SAMPLE`/`stackBucketOf`/`tableBucketOf`, `Family`).
- Dropped a dead `Layout` import and a no-op cast in `SituationCharts.tsx`.
- `noCards` named only one of its two causes in the caption; reworded to cover both.
- `DEFAULT_SCOPE`'s `findIndex` now pinned with an explicit `>= 0` assertion in the test, so a `-1` cannot degrade to a silently-empty chart with a mismatched dropdown.

---

## Not fixed, with reasons

Nothing. All findings were actioned.

The reviewer explicitly confirmed and I want to record it: **`value={scope.heroOffset ?? 'any'}` is correct** — `||` would have collapsed BTN (offset `0`) into "any", `??` does not, and `Number("0")`/`Number("-5")` both round-trip. That was the trap in this PR and it was avoided.

---

## State

- **192 tests** pass (188 → 192), `npm run build` green, new files lint clean (the 7 remaining eslint errors are pre-existing on `master`).
- `classOfHand` re-verified: all **1326** combinations partition into exactly one of the now-fifteen classes, no class unreachable, and no emitted class unlisted.
- Re-rendered against the real 72,840-hand corpus.

McDic:
## Code review iteration 2

Reviewed the delta since round 1 (`ff47197..HEAD` — the EXCLUSIONS mechanism and the label changes). **1 HIGH, 3 MEDIUM, 4 LOW.** All fixed in 3ceb16c.

The HIGH was a spec failure, not a quality nit, and it's the kind I most wanted to avoid: the domain fix that motivated this entire delta was enforced in one chart and quietly violated by the chart directly beneath it.

---

### Issue 1 (HIGH) — The exclusion bound on the ledger but not on the hand-class chart ✅ FIXED

`buildHandClassRows` gated on `family.match(s)` alone and never consulted `EXCLUSIONS`. The iso-raise family still exists (correctly — only the BB *seat* is excluded), so the action dropdown still offered it and `SCOPE_POSITION_KEYS` still offered "Big blind".

Two failures, and the second is the worse one:

1. **"Iso-raise" + "Big blind"** rendered a full class breakdown of decisions the ledger three inches above refuses to score — while its own caption explains that scoring them against folding is a category error.
2. **"Iso-raise" + "All positions"** — *the default* — silently averaged the excluded BB hands in with the legitimate ones, with nothing on the page saying so.

On your real data that second one is not hypothetical: **151 of 473 iso-raises are from the BB — 32% of the bucket.** The pooled row was substantially made of the hands we had just declared uninterpretable.

Both charts now apply the same gate, and the class chart reports the exclusion in the same words the ledger uses. An unexplained blank chart is the same silent absence this whole mechanism exists to prevent, so "Iso-raise + BB" now renders empty *with the reason*:

> Decisions in this action and position: 151.
> Not shown (151): iso-raises from the big blind. In a limped pot the big blind chooses between raising and checking for free — never between raising and folding…

### Issue 2 (MEDIUM) — Nothing tested the ordering the whole mechanism rests on ✅ FIXED

Exclusions are checked *before* families — that ordering is what lets a single seat be carved out of an otherwise-sound family. The reviewer flipped the two and **all 178 tests still passed.**

The reason is instructive: every existing test interrogates the `FAMILIES` and `EXCLUSIONS` *tables* directly; none went through `buildLedgerRows` with an exclusion that *overlaps* a family — because until this delta, none existed. So the single most fragile new behaviour was the one the tests could not reach.

There is now a test that goes through the builder. **I verified it by re-making the mutation** (hoisting the family lookup above the exclusion check — the natural micro-optimisation, since the family path is the common one): the mutant dies, and it dies on exactly that one test.

### Issue 3 (MEDIUM) — The caption misdescribed connectors ✅ FIXED

It said *"Connector and gapper = no high card"* — but **T9s is a connector**, and `classOfHand` tests gap-before-broadway precisely so that it stays one. The caption contradicted the module's own documented, tested, load-bearing ordering decision, in the exact case that ordering exists to handle. A reader would have gone looking for T9s under "broadway-x". Reworded.

### Issue 4 (MEDIUM) — The `FAMILIES` docstring asserted an invariant that is now false ✅ FIXED

It still claimed *"every non-fold decision is matched by exactly one family"*, twenty lines above the `EXCLUSIONS` block that supersedes it. In a file where the invariant *is* the point, a confidently-wrong comment is worse than none — the next person adding a family would read it and reason from it.

### LOW ✅ FIXED

- The 19-line EXCLUSIONS rationale documented the `Exclusion` *interface* rather than the `EXCLUSIONS` table, so hovering the table showed nothing.
- `OFFSETS` in the test retyped `[-5..2]` instead of deriving from `POSITIONS` — drift risk.
- The coverage test's `if (exclusions.length === 1) continue` skipped the double-family check; a family overlapping `isoRaise` only at the BB would have gone unnoticed until the exclusion was later narrowed. Now rejects a double-match regardless.
- `"Not shown: 1 iso-raises"`. Reworded to `"Not shown (1): …"`, which sidesteps pluralisation — i18next's `_one`/`_other` variants are unusable here anyway, since `callSites.test.ts` would flag them as keys the source never names.

---

## Not fixed

Nothing. All findings actioned.

---

## State

- **182 tests** pass (178 → 182), `npm run build` green, new files lint clean (the 3 remaining eslint errors are pre-existing in `allInEquityAsync.ts` on `master`).
- The ordering guard is **mutation-verified**, not merely present.
- Re-checked against the real 72,840-hand corpus: "Iso-raise · BB" is refused and explained; the pooled view no longer blends it in.

McDic:
## Code review iteration 3

Reviewed the delta since round 2 (`3ceb16c..HEAD` — hover descriptions, the exclusion-caption removal, and the master merge). **No CRITICAL, no HIGH.** Two MEDIUMs and two LOWs, all fixed in a5f7489.

The reviewer also independently confirmed three things I'd claimed, which is worth recording:

- **The ordering guard still kills its mutant** after `excluded` counts were removed from `buildLedgerRows` — they re-applied the mutation and watched the test die.
- **No injection risk in `hoverLabel`.** They scanned every string reaching `hovertext`: no `<`, `>` or `&` anywhere, and the only user-derived value in the whole tooltip is `n`, an integer.
- The ledger/hand-class **asymmetry is coherent**, and nothing was orphaned by the caption removal.

---

### Issue 1 (MEDIUM) — One of my tests was proving nothing ✅ FIXED

The sharpest finding, and it's mine. `'does not chart an excluded decision…'` fed `limped/check` and `fourBetPlus/raise` — but **neither matches any family**, so the fold path drops them whether or not the exclusion gate exists. The reviewer deleted the gate outright and the test stayed green.

Worse, the comment I'd written on it was false:

> `// 40 excluded decisions would clear a threshold of 30, so this is a real assertion: a`
> `// family lookup that ran before the exclusions would produce two more rows here.`

It would produce **zero** more rows. A future maintainer would have read that, trusted the test guarded the exclusion path, and it did not.

It now also feeds `limped/raise` at the **big blind**, which *does* match a family (iso-raise) and therefore can only be absent because an exclusion removed it. Verified by mutation: deleting the gate now kills **two** tests where it killed one.

### Issue 2 (MEDIUM) — The hover tooltip rendered as wide as the chart ✅ FIXED

Plotly breaks hover text **only** on an explicit `<br>` — no width-based wrapping, no max width. So "Call vs. open + caller(s)" (167 chars) hovered as a single **~1100px line**: about the full width of the chart on a desktop, and clipped at the plot edge on anything smaller. The reviewer also noticed every *other* hovertemplate in this codebase already hard-wraps — these twelve were the outliers.

Fixed by wrapping in `hoverLabel`, **not** by seeding `<br>` into the strings: a translator would then have to re-break every sentence they rewrite, and the Korean rendering is a different length anyway. The code knows the budget; the translator should only have to know the language.

Measured in a browser rather than asserted: **373px in a 1384px chart**, across two lines.

New `deltaFigure.test.ts` pins it against the **real English strings**, not a stub — a stub translator returns bare keys, which contain no spaces at all and would happily "prove" that an unwrappable string wraps.

### Issue 3 (LOW) — `summarize` was spread-fragile ✅ FIXED

Typed `Omit<DeltaRow, 'label'>`, which since the hover change *includes* `description` — and every caller spreads it over a row literal that has already set one. The day it returned an explicit `description: undefined`, the spread would erase it and **every row would silently lose its hover text**. Now a `Pick`, which cannot, and which says the truer thing: summarising a sample has nothing to do with naming it.

### Issue 4 (LOW) — Stale comment + orphaned rationale ✅ FIXED

`handClassProfit` claimed it explains exclusions *"in the same words the ledger uses"* — the ledger no longer uses any, which was the entire point of the previous commit. And the rationale for *composing* that caption (why the reason keys carry no `{{n}}` — `callSites.test.ts` can't vouch for a placeholder at a non-`t()` call site) went down with the ledger's loop instead of moving to the call site that still does it.

---

## Not fixed

Nothing. All findings actioned.

---

## State

- **189 tests** pass (183 → 189), `npm run build` green, new files lint clean.
- Both exclusion guards are **mutation-verified**, not merely present: deleting the gate kills two tests; hoisting the family lookup above it kills one.
- CI is green — the earlier failure was the unrelated wasm-pack downgrade, fixed and merged in #35.

PR #36, opened 2026-07-13
Korean translation for the situation charts, and live filters in the exported HTML

McDic:
The two follow-ups deferred from #33/#34.

## 1. Filters in the exported HTML

On screen the filter bar sits above the charts and says what it is set to. A downloaded file has no filter bar, so the figure arrived with **no record of the data behind it** — and these filters change the answer, not the view. "Open raise · SB" is a different claim at 12bb than at 60bb, and heads-up hands answer a different question than full-ring ones.

Each exported situation chart now carries a provenance line, rendered above the caption and styled apart from it:

> **Filters** — Opponent's raise from: Any · Starting stack: 15–25bb · Players dealt in: Any · Minimum sample: 30

**Every filter is named, including the ones left at "Any".** Listing only the narrowed ones would leave an unfiltered export saying nothing at all — which is also what a file exported *before* this PR says, so a reader could not tell "no filters were applied" apart from "this file does not record its filters". The hand-class chart's scope (which action, from which position) needs no line: it is already in that figure's title.

The label table is a `Record<keyof SituationFilters, TranslationKey>`, so a fifth filter cannot be added without the compiler demanding a name for it. The test cross-checks that against the *runtime shape* of a filters object, which is what would catch a re-implementation that hardcoded the four we have today.

## 2. Korean

94 keys — the ledger, the hand-class chart, both captions, the filter bar, and every family and exclusion description. `ko.json` is now at 100% of `en.json`. Poker terms take the loanwords the Korean-speaking game actually uses (3벳, 스퀴즈, 아이소 레이즈, 커넥터, 갭퍼), and 잡패 for trash.

### The bug translation would have shipped

Plotly breaks hover text **only** on an explicit `<br>`, so `deltaFigure` wraps the family descriptions itself — budgeting lines by `String.length`. That budget is in the wrong unit the moment the text is not English: **a Hangul syllable renders about twice as wide as a Latin character**, so 64 characters of Korean is ~128 characters of English. Translating these strings would have silently re-created the chart-wide tooltip the wrap was written to prevent (the round-3 fix on #34) — a bug introduced *by the act of translating*, in a language the wrap's author never tested.

The ruler now counts rendered width, not characters.

### Why the test uses a second ruler

The wrap test measures lines with an **independent** implementation of display width, not with `displayWidth` itself. A test that used `displayWidth` would hold the same ruler at both ends: break it, and the lines get wider by exactly the amount the assertion stops being able to see.

This is not hypothetical — I built it that way first and mutation-tested it. With `displayWidth` on both sides, stubbing it back to `.length` left **every wrap test green**. With the independent ruler, the same stub fails `hoverLabel (ko)` and correctly leaves `hoverLabel (en)` passing.

## Verification

- 201 tests pass; clean build; lint unchanged (same 7 pre-existing errors as `master`).
- `locales.test.ts` already guards `{{placeholders}}`, Plotly `%{fields}` and markup across languages — so the hovertemplates and `<extra></extra>` are checked automatically.
- Rendered both figures in a real browser with the real Korean strings: widest hover line is **63 columns against a 64 budget**, and the longest row label (`림퍼 상대 아이소 레이즈 · UTG+1`) fits its margin.


McDic:
## Code review — round 1

The reviewer mutation-tested my claims instead of taking them on faith, and found the architecture sound but **three of the four tests written to pin it were hollow**. That is the right thing to have caught. Fixed in `f4787d9`.

### Fixed

**Issue 1 (HIGH) — `rolldown` was a phantom dependency.** `vite.config.ts` imported it, but it wasn't in `package.json`; it resolved only because npm hoists it out of Vite's tree. Under pnpm's nested layout `vite.config.ts` fails to load and **nothing builds**, including the Pages deploy. And `vite: "^8.0.16"` could swap its internal bundler across a breaking major with no signal in this repo. My comment — *"costs no new dependency"* — was precisely the bug. Declared and pinned to `1.0.3`.

**Issue 2 (MEDIUM) — the dev server would have served a stale runtime.** `addWatchFile()` watched only the entry, but the bundle spans **nine** modules — the aggregation itself. Editing `situationLedger.ts` hot-reloaded the app while leaving the virtual module cached, so an export taken from that session would ship a *stale copy of the aggregation*. That's the exact app/export drift this design exists to prevent, reintroduced in the one environment anyone iterates in. Now iterates `bundle.watchFiles`.

**Issue 3 (MEDIUM) — the fold-drop test was a tautology.** It asserted `exportableSituations(folds) === []`, which is the *definition* of the function, not a fact about the charts. The reviewer proved it by adding a fold-matching family to `FAMILIES`: the test stayed green while the export silently lost 78% of its data. It now builds both figures with and without the folds and compares them whole. **Re-ran that same mutation: it now fails, and so does the conformance test.**

**Issue 4 (MEDIUM) — the conformance test compared `traces` but not `caption`.** The caption is where `hidden`, `inScope`, `noCards` and the exclusion tallies live — every number a fold-drop could disturb without moving a single bar. Captions are now read back out of the DOM and compared against the app's.

**Issue 5 (MEDIUM) — the export's live branch had no tests.** The one path shipping executable code *and* user-derived data in the same document. Six cases added: the runtime is inlined and mounted; it's left out when unused; static sections still draw beside it; and neither a hostile card string nor a hostile translation string can break out of the inline `<script>`.

**Issue 6 (LOW)** — the runtime bundle is now rejected at build time if it contains `import.meta` or `__GIT_`. Both break only inside the exported page, and only at runtime.

**Issue 7 (LOW)** — a failed `Plotly.react` in the download now says so instead of leaving a blank white box. #32 made chart failures visible everywhere except, until now, the export.

**Issue 9 (LOW)** — the wire codes now derive from a list rather than an inverted `Record`. A duplicate or a gap decoded to `undefined`, which matches no family — so those decisions would vanish from the exported charts and from nothing else.

**Issue 10 (LOW)** — `allIn` is now packed and round-tripped, and pinned through the charts (an open jam and an open raise differ *only* in that boolean, and they are different rows).

**Issue 11 (LOW)** — `wrap()` now tests `line`, not `lineWidth`, in both branches.

**Issue 12 (LOW) — and this one was a real bug, not just a disagreement.** The reviewer noticed the impl's ruler and the test's ruler covered different jamo ranges. Chasing it, my first "fix" made it worse: I counted *all* jamo as full-width. But combining jamo (medial/final) render **into** the preceding syllable block, so a decomposed `오` is two columns, not four. Counting each at 2 would have wrapped a decomposed line half as wide as it should be — the `String.length` bug from the other direction. Combining jamo are now zero-width, and a new test pins the two rulers together so "independent" cannot quietly become "different".

### Not fixed, with reasons

**Issue 8 — `<\/script>` → `</script>` in the source template literals.** Deliberate, and it isn't defence being removed. Those are `.ts` template literals, not inline HTML: `\/` in a template literal *is* `/`, so the emitted bytes are identical either way, and ESLint flags it as `no-useless-escape` (it was 2 of the 7 pre-existing lint errors on master). The escape was cargo-culted from writing `</script>` inside an actual inline script. The real defence is `embedScript()`, which this PR adds, and which is now tested.

**i18n — `수티드` / `오프수트` romanize the same root two ways.** They do, and that's the idiomatic pair in Korean poker usage — "suited" and "off-suit" are borrowed as separate words, not as a consistent transliteration of one root. @McDic, shout if you'd rather have them regularized.

**i18n — ko uses `'...'` where en uses `"..."`.** Cosmetic, and it keeps the JSON free of escaped quotes.

### Verification

228 tests pass; clean build; lint **5 errors, down from 7 on master** (added none, fixed two).

PR #38, opened 2026-07-18
Add Deep Dive tab with final-table analysis

McDic:
## What

A new **Deep Dive** tab — the first analysis that joins tournament summaries with hand histories — with its first section, **Final Table Runs**. For every final table the Hero reached, it shows the entry state (rank + chip share, from the hand history) next to the outcome (finish + field size, from the summary), matched on the exact `Tournament #<id>` key.

## Why joined, not hand-history-alone

A single player's GG hand history can't reliably identify the final table: no final-table marker, no placement, no remaining-player count, and seat-count quirks (8→9 expansion, 2-max heads-up spin-off tables) are format-dependent. The tournament summary supplies the exact finish + entrants; joining on `Tournament #` (present in both models) is deterministic, not fuzzy.

## How it works

- **Detection rule:** `finish ≤ maxSeats(last table played)` — a finish of rank R means only R players remained, so R ≤ table capacity means one table was left (the final table). Robust to simultaneous bust-outs and the heads-up spin-off (folded back into the final table for the entry hand).
- **Parser:** now keeps `HandHistory.tableId` (the regex already matched it, then discarded it) to group hands into the final-table run.
- **Filtering:** excludes satellites/qualifiers (name has a whole-word `to` / `Satellite` / `flip stage`) and micro-fields (< 10 entrants). Verified across all 749 distinct tournament names that whole-word "to" appears only in satellites; real events like "Flip & Go [Go Stage]" and standalone Flipout cash are kept. Excluded tournaments are recorded with a reason, never silently dropped.
- **UI:** the tab is disabled until both datasets load (consistent with the other tabs); the table has sortable columns (default date-descending) and a note that the list is joined data and may omit final tables from months without downloaded hand histories.

## Verification

- 241 unit tests pass (incl. i18n parity/call-site checks and component render tests for the tab, table, and sorting).
- `npm run build` typechecks and builds.
- Ran the real parsers + analysis over the full `data/damavaco` dataset: the list resolves to the 3 real MTT final tables (Monster Stack 7/2695, Crazy Eights 8/1462, Fifty Stack 1/2360), with 53 satellites/micro-fields excluded. HH-derived finishes match the summaries exactly.

## Deferred follow-ups

- Hover tooltip on the disabled tab explaining why it's inert.
- Allow uploading files during the bankroll simulation (the uploader is disabled while `isLoading`, which stays true through the auto-analysis after a tournament upload).


McDic:
## Code review (automated) + fixes

An automated review pass found **no Critical/High issues** (verdict: approve). It confirmed the join is a genuine exact-key inner join, the `finish ≤ maxSeats(last table)` rule is sound, re-entry lumping is correct, chronological re-sorting works, and the sort math is right. Findings and their disposition below — fixes in commit 3db7885.

### Fixed

**Issue 1 (Medium) — skipped tournaments were computed but never shown.** `analyzeFinalTables` tracks a `skipped` list with reasons, but the UI discarded it, so an empty list couldn't be told apart from "everything got filtered." The tab now shows a "*N joined tournaments aren't listed (satellites, fields under 10, or busted before the final table)*" line in both the empty state and below the table.

**Issue 2 (Low) — heads-up fold could discard the real final table.** The 2-max spin-off fold blindly took the previous run. It's now guarded on the previous run being larger than 2-max, so back-to-back 2-max runs or a truncated download can't silently drop the final table.

**Issue 3 (Low) — sort defaults were worst-first for Finish / Entry rank.** First click on those columns now defaults to ascending (1st place / chip leader first); field size and chip share stay high-first.

**Issue 4 (Low) — entrant counts used the browser locale.** Now formatted with the active i18n language.

**Issue 5 (Low) — entry-rank tie semantics undocumented/untested.** Added a comment (competition ranking: ties share the best rank) and a tie test.

**Issue 6 (Low) — test-coverage gaps.** Added tests for out-of-order hands, the unknown-capacity (`maxSeats` sentinel) skip, stack ties, and null `tournamentId`.

**Issue 7 (Medium, partial) — `isSatelliteName` is a name-only, empirically-validated heuristic.** Added "qualifier" to the matched words and documented it as best-effort. Rather than a larger structural change (e.g. prize-shape detection), this is mitigated by Issue 1: any mismatch now surfaces in the skipped count instead of being invisible.

**Issue 8 (Low) — a truncated download containing only the heads-up hands reports a 2-seat "final table."** Left as-is: the player did reach the final table (finished 1st/2nd), and skipping it would drop a legitimate deep finish. Entry metrics over just the two heads-up stacks in that rare truncation case are acceptable; the reviewer also marked this "none required."

**Issue 9 (Low) — two defensive branches remain untested** (the `total === 0` ratio guard and the heads-up fold when only one run exists). The `total === 0` branch is unreachable with real data — Hero always holds > 0 chips at entry, so the table total is > 0 — so it's a pure guard; not worth contriving impossible inputs for.

All 245 tests pass; `npm run build` typechecks and builds.

PR #39, opened 2026-07-18
Disabled-tab tooltips; parse in its own worker (Deep Dive follow-ups)

McDic:
Two follow-ups to the Deep Dive PR.

## Item 1 — disabled-tab tooltips (all tabs)

Every disabled tab now shows a hover tooltip explaining which data it needs — Tournament → "Needs tournament results", Hand History / Preflop Situations → "Needs hand history data", Deep Dive → "Needs both tournament results and hand histories". `ChartTabs` is refactored to be config-driven so the four tabs share one implementation.

It's a CSS `::after` tooltip (not a native `title`, which doesn't reliably fire on a disabled `<button>`), positioned **above** the tab with a down-arrow so it no longer covers the first chart section.

## Item 2 — upload during the bankroll simulation

**Parsing and the bankroll simulation now run on separate workers** (each with its own WASM instance). Previously they shared one worker, so a hand-history upload during the sim was queued behind it and its charts/all-in-equity couldn't start until the sim finished. Now the parse runs concurrently and the hand-history analysis starts right away — matching how the all-in equity worker already behaves.

The bankroll-sim **progress moved below the tabs**, into the tournament tab's loading slot, showing the real "Simulating N buy-ins…" message — consistent with the equity progress in the hand-history tab. The generic above-tabs "Analyzing…" indicator is removed.

Memory cost of the second worker: ~5–20 MB (a lightly-used WASM instance; it does **not** load the 686 KB preflop cache, which parsing never touches).

## Verification

- 247 unit/render tests pass, incl. new tests for the all-tabs tooltips and the below-tabs sim message; `tsc -b` and `npm run build` are clean.
- Interactive behavior (concurrent parse during the sim; tooltip positioning) confirmed via `npm run dev`.

### DIFF
--- a/web/src/App.css
+++ b/web/src/App.css
@@ -260,6 +260,48 @@
   cursor: not-allowed;
 }
 
+/* Tooltip explaining why a tab is inert. A CSS `::after` rather than a native `title`, which does
+   not reliably show on a disabled <button>; `:hover` still matches the disabled tab. */
+.chart-tabs .tab[data-tooltip] {
+  position: relative;
+}
+
+.chart-tabs .tab[data-tooltip]:hover::after {
+  content: attr(data-tooltip);
+  position: absolute;
+  /* Above the tab, so it never covers the chart section below. */
+  bottom: calc(100% + 10px);
+  left: 50%;
+  transform: translateX(-50%);
+  width: max-content;
+  max-width: 220px;
+  white-space: normal;
+  text-align: center;
+  background: #333;
+  color: #eee;
+  padding: 0.45rem 0.7rem;
+  border-radius: 6px;
+  font-size: 0.8rem;
+  font-weight: 400;
+  line-height: 1.35;
+  z-index: 100;
+  pointer-events: none;
+  box-shadow: 0 4px 14px rgba(0, 0, 0, 0.55);
+}
+
+/* Arrow under the bubble, pointing down at the tab. */
+.chart-tabs .tab[data-tooltip]:hover::before {
+  content: '';
+  position: absolute;
+  bottom: calc(100% + 4px);
+  left: 50%;
+  transform: translateX(-50%);
+  border: 6px solid transparent;
+  border-top-color: #333;
+  z-index: 100;
+  pointer-events: none;
+}
+
 .chart-tabs .tab-icon {
   font-size: 1.25rem;
 }
--- a/web/src/App.css
+++ b/web/src/App.css
@@ -266,7 +266,8 @@
   position: relative;
 }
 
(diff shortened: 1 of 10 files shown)


McDic:
## Code review (automated) + fixes

An automated review pass **approved** with no Critical/High issues. It specifically tried to break the two-worker coordination and confirmed it correct: `tournamentsRef` is current when a queued analyze posts (the hook's sync effect runs before App's `runAnalysis` effect), no missed/duplicate analyze across interleavings, parse/analyze results touch disjoint state via functional `setState`, each worker has its own WASM memory so concurrency is safe, and the parse queue never loses work. Findings and dispositions below (fixes in commit 24d6c1d).

### Fixed

**Issue 1 (Medium) — disabled-tab tooltip was mouse-only.** The CSS `:hover` tooltip gave keyboard and screen-reader users no way to learn why a tab is inert. Disabled tabs now use `aria-disabled` instead of the `disabled` attribute (so they stay focusable), carry an `aria-label` with the reason, and the tooltip also shows on `:focus-visible`.

**Issue 2 (Low) — effect cleanup left the coordination refs set.** The worker-init effect's cleanup now resets `parseBusyRef`/`analyzeBusyRef`/`pendingFilesRef`/`reanalyzePendingRef`, so a real mid-work remount can't leave a pump wedged against a terminated worker. (StrictMode's init-time remount has nothing in flight, so this is insurance.)

**Issue 3 (Low/Medium) — two WASM instances coexist.** Intentional: the whole point of the two-worker split is to parse concurrently with the bankroll sim, and separate WASM memory is what makes that safe. Measured overhead is ~5–20 MB (the parse worker never loads the 686 KB preflop cache), which is acceptable on the target machines.

**Issue 4 (Low) — `reset()` doesn't fence in-flight tasks.** `reset` is returned from the hook but not wired up anywhere in the app, so this is dead/latent. If it's ever used, the fix is to bump a generation counter that result handlers check (or terminate+recreate the workers); noted for then rather than adding unused code now.

**Issue 5 (Low) — pump closures are assigned to refs during render.** Benign "latest ref" pattern: both closures read only refs and the stable `setState`, so every render's closure is behavior-equivalent and a discarded concurrent render can't cause a stale-behavior call. Left as-is to avoid the dep-management complexity of converting to `useCallback`/effect.

**Issue 6 (Low) — starting a parse clears the whole `errors` array.** This matches the pre-existing single-worker behavior (`parseFiles` always cleared errors), so it's not a regression. Partitioning errors by source is a reasonable future refinement but out of scope here.

All 247 tests pass (incl. the updated a11y assertions); `tsc -b` and `npm run build` are clean.

PR #40, opened 2026-07-18
Render heavy charts with WebGL (scattergl): RR-by-Rank + Chip Histories

McDic:
## What

Move the two heaviest scatter charts from SVG to Plotly's WebGL renderer (`scattergl`), where the point/line count makes SVG janky on hover and zoom.

### RR-by-Rank
One marker per cashed tournament — up to a few thousand across a full 12-month history (3,769 summaries in the real damavaco data before the cash filter). Switches the two per-tournament marker traces (main RR + the legend-only PERR trace) to `scattergl`. The regression line stays SVG `scatter`: a handful of fitted points (no perf gain from gl), and as SVG it paints *above* the WebGL layer, keeping the trend line on top of the dots.

### Chip Histories
The main panel draws one line per tournament — ~943 lines / ~76,100 vertices on real data, all re-projected through a log-y axis every pan/zoom frame. Switches just the per-tournament trajectory lines to `scattergl`. The three one-off overlays stay non-gl on purpose: the dashed danger line and the death-threshold bar are one trace each (no perf reason), the survival curve needs `fill` (gl's weak spot), and being SVG they paint above the WebGL layer so the danger line and overlays stay on top.

## Why it's safe

- **Export unaffected.** The standalone HTML export re-embeds trace JSON and re-renders against the full Plotly CDN build (`plotly-3.3.1.min.js`), which ships the gl renderer. Verified the exported Chip Histories section renders.
- **WebGL context budget.** These two charts live in separate graph divs → 2 contexts, well within the browser ceiling, even with all tabs kept mounted (`display:none`).
- **Fidelity verified in the browser on real data** (Chip Histories): appreciably smoother hover/zoom once loaded (slightly longer initial load, acceptable); log-y axis, dashed danger line, other panels, annotations, and HTML export all correct.

## Tests

- New intent-pinning tests: RR-by-Rank markers are `scattergl` / trend line stays `scatter`; Chip Histories per-tournament lines are `scattergl` while the danger line / survival curve / death-threshold overlays stay non-gl (identified by name, since they share `mode:'lines'`).
- Full web suite green (249 tests); `npm run build` typechecks clean.

### DIFF
--- a/web/src/App.tsx
+++ b/web/src/App.tsx
@@ -120,6 +120,17 @@ function App() {
     }
   }, [tournaments, runAnalysis])
 
+  // Plotly sizes each chart to its container at draw time. A chart first drawn inside a hidden
+  // (`display:none`) tab is drawn at width 0: an SVG chart still recovers, because the browser
+  // scales its viewBox to the container once shown, but a WebGL (`scattergl`) chart's drawing
+  // buffer stays 0×0 and paints blank until something resizes it. react-plotly's `useResizeHandler`
+  // listens on window resize only, so nudge it on every tab switch — by the time this effect runs
+  // the now-visible tab has its real width, and the dispatch makes every mounted Plot re-fit. This
+  // became load-bearing when the RR-by-Rank and Chip Histories charts moved to WebGL.
+  useEffect(() => {
+    window.dispatchEvent(new Event('resize'))
+  }, [activeTab])
+
   // The export follows whatever language the site is currently in: the charts were
   // built in it, so anything else would mean rebuilding every figure.
   const language = i18n.resolvedLanguage ?? 'en'
--- a/web/src/visualization/handHistory/chipHistories.ts
+++ b/web/src/visualization/handHistory/chipHistories.ts
@@ -98,8 +98,9 @@ export async function getChipHistoriesData(
     // One gl line per tournament. This is the heavy trace — ~943 lines / ~76k vertices on the
     // real data — and it is the only thing here worth moving to WebGL. The danger line, survival
     // curve, and death-threshold bar below stay SVG on purpose: they are one trace each (no perf
-    // reason), the survival curve needs `fill` (gl's weak spot), and being SVG they paint above
-    // this WebGL layer so the danger line and overlays stay on top of the chip trajectories.
+    // reason) and the survival curve needs `fill` (gl's weak spot). The danger line also shares
+    // this subplot with the chip lines, and being SVG it paints above the WebGL layer, so it stays
+    // on top of the trajectories (the survival curve and bar sit on their own grid panels).
     traces.push({
       type: 'scattergl',
       x: xValues,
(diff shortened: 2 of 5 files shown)


McDic:
## Code review disposition (`/cri`)

A code-reviewer agent reviewed the diff (both source edits, both new tests, the export path, the render/tab-mount setup, and the two deliberately-not-converted siblings). Verdict: **approve — 0 Critical, 0 High, 1 Medium, 2 Low.** Details below.

### Fixed

**Issue 1 (Medium) — WebGL chart first-painted inside a hidden tab can render blank.** All tabs stay mounted via `display:none`, and a `scattergl` chart drawn at container width 0 keeps a 0×0 drawing buffer and paints blank until something resizes it — unlike an SVG chart, whose viewBox the browser rescales to the container when shown. react-plotly's `useResizeHandler` listens on window resize only, and there was no resize-on-tab-activation anywhere. Because the default tab is Tournament and Chip Histories computes asynchronously, a user who waits on the Tournament tab until analysis finishes and *then* opens Hand History could hit this.
Fix (commit 7cede96): dispatch a `window` resize on every `activeTab` change — by the time the effect runs the now-visible tab has its real width, so every mounted Plot re-fits (a no-op for already-correct SVG charts). This is the defensive fix for the whole class of bug, not just the two converted charts.

**Issue 3 (Low) — z-order comment overreached.** The "SVG paints above the gl layer so overlays stay on top" note only truly applies to the danger line, which shares the subplot with the chip lines; the survival curve and death-threshold bar sit on their own grid panels and never overlap. Tightened the comment (and its mirror in the test) accordingly.

**Issue 2 (Low, theoretical) — PERR gl markers on an overlaying secondary log axis.** The reviewer noted gl traces on an `overlaying` secondary axis have historically been a glitch-prone Plotly corner, but had no evidence it's broken in 3.3.1, and every property in play (`hovertemplate`, `customdata`, `marker.color`, log axes, `legendonly`) is supported under `scattergl`. The trace is `visible: 'legendonly'`, so any glitch would only surface when a user toggles PERR on. Leaving as-is; worth one manual toggle to eyeball, but not a code defect.

### Confirmed correct (no action)
- **Export** re-embeds trace JSON verbatim and renders against the full Plotly CDN build (`plotly-3.3.1.min.js`), which ships the gl renderer; `type:'scattergl'` is a plain JSON string and survives serialization.
- **WebGL context budget**: these are the only gl traces in the codebase → exactly 2 contexts (separate graph divs), far under the browser ceiling; all-tabs-mounted doesn't change that.
- **Scoping**: the not-converted siblings are correctly excluded — `deltaFigure` uses `error_x` error bars (unsupported by `scattergl`) with a small bounded point count, and `historicalPerformance` is ~6 line traces (cheap SVG paths; the perf pain is thousands of markers/lines, which is what got converted).

### Verification
Full web suite green (249 tests); `npm run build` typechecks clean.

One manual check still worth doing before merge: upload data, stay on the Tournament tab until all analysis finishes, then switch to Hand History — confirm the Chip Histories trajectories render immediately (this is the exact scenario Issue 1 guards against).

PR #41, opened 2026-07-18
Add explanatory captions to every tournament & hand-history chart

McDic:
## What

The Tournament and Hand History charts had no on-screen documentation — only the Deep Dive and Preflop Situation tabs did. This adds a caption to each of the **eight** older charts, in **both English and Korean**:

- **Tournament**: Historical Performance, RRE Distribution, Bankroll Analysis, Prize Distribution, RR by Rank
- **Hand History**: Chip Histories, Hand Usage (VPIP) Heatmaps, All-in Equity

## How

Each visualization module now builds a `caption: string[]` (one paragraph per line), exactly like the situation figures already do. Because `getChartData` spreads the figure into `ExportChart`, and `ExportChart` already renders a `caption`, the same text flows to **both the in-app render and the HTML export with no new export wiring**. A small shared `ChartCaption` component lays the lines out above each figure, mirroring the export.

Captions are present with data **and** on each chart's no-data path (the caption is where "what am I looking at" lives even when the chart is empty).

## Content notes (verified against the code)

- **Prize Distribution**: the two rings aren't a like-for-like split — the weekday ring covers every prize, but only tournaments above 0.5% of winnings get their own outer wedge (vs the pie's 1% "Others" cut).
- **Hand Usage**: seats earlier than the 8-max UTG (9-max UTG included) fold into the UTG panel — confirmed by `mappedOffset = offset <= -5 ? -5 : offset`.
- **All-in Equity Luck Score**: described as a Gaussian z-score of your actual win count vs a Poisson-binomial of your per-hand equities — confirmed against `crates/core/src/equity.rs` (`luck_score` = `Φ⁻¹(lower-tail p-value)`).
- Korean uses the established terminology (수딧 / 옵숫, 멘징, 딥런, 행운 지수, 뱅크롤 …).

## Tests

- Each chart builder is asserted to return a non-empty caption, both with data and on the no-data path.
- i18n parity (`locales.test.ts` / `callSites.test.ts`) green — every new key has a literal call site and en/ko match (38 caption keys each).
- Full web suite green (257 tests); `npm run build` typechecks.


McDic:
## Code review disposition (`/cri`)

A code-reviewer agent verified the wiring, every return path, i18n parity, and — importantly — each caption's factual claims against the actual source (and the Rust `equity.rs` for the Luck Score). Verdict: **one HIGH inaccuracy, everything else correct.**

### Fixed

**Issue 1 (High) — RR-by-Rank caption had the spatial claim inverted.** It said deep, high-return runs sit in the "lower-left"; they're actually in the **upper-right**. The percentile x-axis is reversed (`range: [0, log10(minPercentile) − 0.2]` — a negative upper bound), so deeper finishes (smaller percentile) are on the right, and higher RR is on top. "Lower-left" is exactly where shallow min-cashes cluster, so a reader following the caption would read the chart backwards. The user independently caught the same thing.
Fix (commit 54a6348): corrected to "upper-right" in both locales, and added a note that the percentile axis is reversed (that inversion is easy to misread).

### Verified correct — no action

- **Wiring**: all 8 charts spread the full figure in `getChartData`, so `caption` reaches both the in-app render and the HTML export (`htmlExport.ts` renders `chart.caption ?? []`). No chart drops it.
- **Every return path**: caption is present on both the main and no-data returns of the four modules with an early return, and on the single return of the four inline ones. `caption` is required in the interfaces, so the passing typecheck guarantees no `undefined.map` crash in `ChartCaption`.
- **i18n parity**: all 23 new keys exist in both locales with exactly one literal call site each — no orphans either direction.
- **Other content claims all match the code**: Hand Usage 9-max-UTG folding (`offset <= -5 ? -5 : offset`); Prize ring 0.5% outer-wedge vs 1% "Others" mismatch; Historical "profitable ratio" = profit > 0 (prize above buy-in); All-in Luck Score = `Φ⁻¹(lower-tail)` of a Poisson-binomial of your equities (a Gaussian z-score).

### Verification
JSON parses; i18n parity + tournament tests green; full suite was green at 257; `npm run build` typechecks.

PR #43, opened 2026-07-26
Count covering calls of all-ins as all-in showdowns in equity chart

McDic:
## Summary

The All-in Equity chart's eligibility check required Hero's own action line to carry GG's `and is all-in` tag. This missed two whole classes of genuine all-in showdowns:

- **Covering calls** — Hero calls a shove with the bigger stack. GG prints no all-in tag for the caller, so the hand was silently dropped (e.g. HU: opponent shoves 360, Hero's 540-stack calls; board runs out — a textbook all-in showdown).
- **Blind/ante all-ins** — the parser's postprocess flags short blind/ante posts with `isAllIn: true` on the action, but never records them in `h.allIned`, so they were invisible to eligibility too.

## Change

New helper `getHandHistoryAllInShowdownStreet(h, playerId)` in `web/src/types/index.ts` qualifies a hand when:

- the player went all-in themselves (tagged action, or all-in from posting a blind/ante), returning their all-in street; or
- at the end of some street, every other remaining player is all-in while the player covers — betting is closed at that point, so the board runs out exactly as if the player were all-in too.

Both eligibility call sites (`allInEquityAsync.ts`, `analysisWorker.ts`) now use it. The old one-line accessor `getHandHistoryAllInedStreet` had no other consumers and was removed. River all-in showdowns stay excluded as before (no board left to run out).

## Validation on real data (76,351 hands)

| | hands |
|---|---|
| Eligible before | 2,623 |
| Eligible after | **3,898 (+49%)** |
| … from covering calls | +1,247 |
| … from Hero blind/ante all-ins | +19 |
| … covering calls over blind/ante all-in opponents | +9 |

Per-street split after: 3,127 preflop / 419 flop / 352 turn. The counts reconcile exactly with an independent classification of all showdown hands.

## Tests

`web/src/types/allInShowdown.test.ts` drives the detection from raw GG hand text (so the parser's tagging, and its absence on covering calls, is part of what's tested): the five discussed scenarios, plus turn-covering-call street selection, regular showdown rejection, Hero blind all-in, and Hero folding to a shove. `npm test`: 271 passed.

https://claude.ai/code/session_01K5UHXgXJq8A2A9X7Qm3YLf


McDic:
## Code Review (round 1, automated reviewer)

### Correctness

**[minor] Expanded reliance on the ante/blind all-in heuristic is untested for ante and unvalidated for BB-ante formats**
`web/src/types/index.ts` — `getPlayerAllInStreet` treats any `blind`/`ante` action with `isAllIn` as a genuine all-in, both for Hero's own eligibility and for classifying opponents in the covering loop. That flag comes from `postprocessHandHistory`'s heuristic (`amount < maxAnte` / `bb > amount`). In a BB-ante format (only BB posts the ante), short/zero antes from other players could be mis-flagged as all-in, producing false-positive eligible hands.

**[minor] Showdown-player set can diverge from the all-in street's opponent set (pre-existing, now more exercised)**
When Hero is all-in early but a side pot continues and an opponent folds later, equity is computed vs fewer opponents than Hero was actually all-in against. Unchanged behavior, but the PR broadens how often such hands qualify.

### Performance

**[medium] Per-hand cost regressed from O(1) to O(streets × players × preflop-actions) on the main thread over ~76k hands**
The old accessor was a single `Map.get`. The new function iterates all four streets, rebuilds `remaining` each street, and calls `getPlayerAllInStreet` (which re-scans all of `actionsPreflop` for any untagged player) up to 4 × remaining times per hand — for the majority of hands that have no all-in at all. Suggested: early-return when the hand has no all-in signal; precompute per-player all-in streets once.

**[minor] Eligibility recomputed per hand in `analysisWorker`**
`getHandHistoryAllInShowdownStreet(h, 'Hero')` is called twice per eligible hand (filter, then loop). Compute once and carry the result.

### Test coverage

- **[minor]** The "+9 covering-call-over-a-blind/ante-all-in-opponent" real-data category is not tested (opponent all-in only via the postprocess flag, no GG tag).
- **[minor]** Ante all-in path entirely untested (all scenarios use blinds).
- **[minor]** No flop covering-call test; no multiway false-positive guard test (one opponent all-in, second opponent live with chips → must return null).
- **[minor]** No helper-level test that a river all-in returns `'river'` (river exclusion lives only in the two callers).

### Nits

- **[nit]** Helper computes the river covering case only for callers to discard it; the "river excluded by design" contract is duplicated across two call sites.

### Summary
No critical or major defects. The two-way qualification logic (own all-in incl. blind/ante; covering call via single-non-all-in) is sound — fold accumulation, per-street all-in ordering, fold guard, and earliest-street selection all check out; reviewer could not construct a false positive or false negative. Recommendation: **APPROVE** with the performance early-out and test gaps addressed.

McDic:
## Review round 1 — resolution (commit 9e6c979)

### Fixed

**Issue 1 (medium, performance): per-hand cost regressed to O(streets × players × preflop-actions).**
Fixed. `getPlayerAllInStreet` (per-player re-scan of `actionsPreflop`) is replaced by `getAllInStreets(h)`, which builds the per-player all-in street map once per hand: a copy of `h.allIned` plus a single pass over preflop blind/ante flags. The function now also returns immediately when the map is empty — the common no-all-in case skips the street loop entirely. Re-validated on the real dataset: identical output (3,898 eligible; 3,127/419/352 by street), full 76,351-hand classification including parsing runs in ~1.8s.

**Issue 2 (minor, performance): eligibility recomputed per hand in `analysisWorker`.**
Fixed. The filter now produces `[hand, street]` pairs and the equity loop consumes the carried street; the duplicate helper call and the non-null assertion are gone.

**Issue 3 (minor, tests): uncovered branches.**
Fixed — five new raw-hand-text tests pin: covering an opponent all-in from a short small blind (postprocess flag only, no GG tag — the "+9" real-data category), Hero all-in from a short ante, a covering call on the flop resolving to 'flop', a second live opponent with chips preventing qualification (false-positive guard), and the helper reporting 'river' for a river covering call (exclusion stays the callers' policy). Suite: 271 passed.

### Not fixed, with reasons

**Issue 4 (minor): blind/ante all-in heuristic could mis-flag in BB-ante formats.**
The heuristic lives in `postprocessHandHistory` and predates this PR; this PR only consumes its output. GG's tournament exports in the real dataset all use per-player antes, and the full-dataset validation reconciled exactly against an independent classification with no anomalous eligible hands. If a BB-ante format ever appears, the fix belongs in the parser postprocess, not in eligibility — noted as a potential follow-up.

**Issue 5 (minor): showdown-player set can diverge from the all-in-street opponent set when a side pot continues and someone folds later.**
Acknowledged pre-existing behavior, deliberately kept: the project's chosen definition is "Hero all-in before the river qualifies", and the folded player's hole cards are never revealed, so computing equity against the eventual showdown hands is the best available estimate. Affects 13 of 3,898 eligible hands in the real dataset.

**Issue 6 (nit): helper computes the river case only for callers to discard it.**
Kept: the helper's contract is "street where the all-in showdown formed"; excluding river is chart policy, and the new river test documents exactly this split. The extra work is a few comparisons on an already-qualified hand.

PR #45, opened 2026-07-31
Fix all 7 lint errors and gate eslint in CI

McDic:
Follow-up to #44, which noted that lint was ungated and 7 errors had accumulated on master. This fixes all 7 and closes the gate behind them.

Three commits, reviewable independently and in order.

## 1. Four mechanical fixes

- `parser/utils.ts` — `\$` inside a character class needs no escape. `[\$¥฿₫₱₩]` and `[$¥฿₫₱₩]` match identically, so this is presentational only.
- `allInEquityAsync.ts` — three `total*` counters were accumulated from every worker's `result` message and then never read; the function returns only `{ data }`. They are the residue of a debug log removed without its feeding code. `msg.stats` stays in the worker protocol (equityWorker still reports it, still useful for ad-hoc profiling) — only the dead consumer goes.

## 2. Three render-purity hazards

These are the ones worth actual review.

**`useAnalysisWorker` — ref writes during render.** The two pumps were assigned into refs on every render. The comment defending it gave the game away:

> Reassigned each render, but closes over only refs/setState

That is precisely the condition under which the "latest ref" pattern is unnecessary. Neither closure captured a single reactive value, so every per-render rebinding produced a behaviourally identical function — the indirection bought nothing and cost a render-phase ref write that is unsafe under concurrent rendering.

They are now plain `useCallback(…, [])`: stable for the life of the component, exactly as the ref indirection intended. The worker-init effect and the two public callbacks list the pumps as dependencies; since the pumps have empty dep arrays, the workers are still created exactly once.

**`App` — cascading render.** The auto-switch-tab effect called `setState` synchronously, so React committed the stale tab, painted it, then immediately re-rendered with the corrected one. Beyond the lint error that is a visible flash of the wrong tab on upload. Moved to a state adjustment during render guarded by the tracked counts (React's documented pattern for this); React now discards the pass and re-runs before committing, so only the corrected tab reaches the DOM. The guard is what prevents a loop — the branch runs only on a render where a count actually changed.

**Behaviour is preserved deliberately.** Note that the old effect would yank the user's tab back whenever a count changed, even if they had manually navigated away. That is arguably a bug, but it is pre-existing and out of scope here; the render-phase version reproduces it exactly.

### New test coverage

The queueing this touches had **no** test — `useAnalysisWorker.test.ts` covers only the exported merge helpers. `useAnalysisWorker.queue.test.ts` drives the hook through a fake `Worker` (house style: raw `createRoot` + `act`, no new dependency) and pins what actually matters:

- an upload arriving mid-parse is queued and posted when the worker frees
- several mid-parse uploads coalesce into one follow-up batch
- the queue drains after a worker `onerror`, so one bad batch cannot wedge uploads
- repeated mid-sim re-analysis requests collapse to a single re-run
- that re-run carries the tournaments merged *since the sim started*, not the ones it began with

Both pump call sites were mutation-tested: deleting either `pump*()` call from its `onmessage` handler fails exactly two tests. The tests have teeth.

## 3. The gate

`npm run lint` added to `test-web.yml`. That workflow rather than `pre-commit.yml` because pre-commit has no Node toolchain, and this one has already run `npm ci` a step earlier — so the gate is one step and a few seconds, not a new job. It needs no type information of its own: the config extends `tseslint.configs.recommended`, not the type-checked variant.

## Verification

| Step | Before | After |
|---|---|---|
| `npm run lint` | 7 errors | **0** |
| `npx tsc -b` | clean | clean |
| `npm test` | 271 passed | **282 passed** (+11) |


McDic:

**Verdict: APPROVE** — no critical or major issues. The reviewer verified the PR's two central claims by direct experiment rather than by reading, which is worth recording.

### Verification performed

**The `useCallback` conversion is behaviour-preserving.** Every identifier each pump closes over was enumerated: `pumpParse` captures `parseWorkerRef`, `parseBusyRef`, `pendingFilesRef`, `pendingAllowFreerollsRef`, `setState`, `STARTING_PROGRESS`; `pumpAnalyze` captures `analyzeWorkerRef`, `analyzeBusyRef`, `reanalyzePendingRef`, `tournamentsRef`, `setState`, `STARTING_PROGRESS`. Zero reactive captures, so the `[]` deps are honest.

**Workers are still created exactly once.** The hook was driven through 16 commits (5 parse round-trips plus progress messages): 2 `Worker` instances, 0 terminations, 1 distinct `parseFiles` identity and 1 distinct `runAnalysis` identity throughout. Identity stability matters beyond this file — `App.tsx:117-121` has `useEffect(..., [tournaments, runAnalysis])`, which would loop if `runAnalysis` were unstable.

**The render-phase state adjustment is correct.** Old and new logic were run side by side through 7 scenarios. Exactly one extra render pass (no loop), correct under StrictMode, correct when the user has manually navigated, and the flash-elimination is observable in the commit trail: `[situation, tournament]` becomes `[tournament]`.

**The new tests have teeth.** Six mutations, all killed:

| mutation | tests failed |
|---|---|
| drop `pumpParse()` in parse `onmessage` | 2 |
| drop `pumpParse()` in parse `onerror` | 1 |
| drop `pumpAnalyze()` in analyze `onmessage` | 2 |
| drop `worker.terminate()` | 1 |
| `pumpAnalyze` captures `state.tournaments` (stale closure) | 3 |
| `pumpParse` deps `[state]` (unstable identity) | 6 |

No false-confidence tests found. The deleted stats aggregation was confirmed dead by grep, and the regex escape confirmed equivalent (`ANY_MONEY` has flag `g` only, no `u`).

### Findings

1. **[MINOR]** `App.tsx:198` — `prevCounts` seeded from live counts, so the auto-switch is skipped if the component ever mounts with data already present. Unreachable today, silent when it stops being so.
2. **[MINOR]** `useAnalysisWorker.queue.test.ts` — `IS_REACT_ACT_ENVIRONMENT` not set, which suppresses React's "update not wrapped in act(...)" warning.
3. **[MINOR]** `useAnalysisWorker.queue.test.ts:219-229` — the teardown test's re-mount sits after the assertions, so a real failure cascades into the shared `afterEach`.
4. **[MINOR]** `useAnalysisWorker.queue.test.ts:59` — `latest.current` is never reset between tests, so the `api()` guard cannot actually fire.
5. **[MINOR]** `test-web.yml` — lint runs after the multi-minute Rust/WASM build it does not need.
6. **[NIT]** Redundant `@vitest-environment` docblock.
7. **[NIT]** Non-lazy `useState` initializer.
8. **[NIT]** Index-based worker accessors couple the tests to construction order.
9. **[NIT]** `allowFreerolls` is never asserted, despite the coalescing path applying the last upload's flag to the whole merged batch.

### Observation (pre-existing)

`useAnalysisWorker.ts:173-175` clears `reanalyzePendingRef` *before* the empty-tournaments early return, so a `runAnalysis()` issued with no tournaments is dropped rather than deferred. Unreachable today; the new test `does not post an analysis when there is nothing to analyse` now characterises it.

On the path filter: **adequate**. `eslint .` runs with cwd `web/` and the flat config only has rules for `**/*.{ts,tsx}`, all of which live under `web/**`.

McDic:
## Response to review — pushed as `02f8da9`

Five of the nine findings fixed, four declined with reasons. Suite is green after the changes: **lint 0 errors, `tsc -b` clean, 283/283 tests** (up from 282 — one new test from Issue 9).

### Fixed

**Issue 1 — `prevCounts` seeding.** Fixed, and the most worthwhile finding here. Seeded with literal zeros instead of the live counts:

```ts
const [prevCounts, setPrevCounts] = useState({ tournaments: 0, handHistories: 0 })
```

The effect this replaced ran on mount; seeding from live counts did not. Today the two are indistinguishable because `useAnalysisWorker` always returns `initialState` on first render, so mount counts are always `0/0`. But the divergence is real and would be silent — a hand-history-only user left staring at an empty Tournament tab — the moment initial state comes from anywhere else. Zero behaviour change today.

**Issue 2 — `IS_REACT_ACT_ENVIRONMENT`.** Added to `beforeEach`, matching `TournamentCharts.test.tsx:45` and `HandHistoryCharts.test.tsx:89`.

One correction to the finding: the stated symptom does not reproduce. The claim was ~40 lines of `not configured to support act(...)` in `npm test` output; the actual count both for the file in isolation and for the full suite is **zero** — the isolated run is 9 lines total. So this was not fixing observable noise. The *other* half of the finding is sound and is why it was applied anyway: setting the flag is what makes "an update was not wrapped in act(...)" an actual warning, which is the net that catches a future `emit()`-style helper that forgets to wrap.

It went in the local `beforeEach` rather than `src/test/setup.ts` as suggested. `setup.ts` would fix all four files at once, but enabling the flag globally also *enables* the unwrapped-update warning for `ChartTabs` and `DeepDiveCharts`, which have never run under it — that is a change to unrelated tests and belongs in its own PR, not smuggled into this one.

**Issue 3 — teardown `finally`.** Fixed; the re-mount now runs in `finally` so a genuine termination regression reports once.

**Issue 4 — `latest.current` reset.** Fixed; cleared in `beforeEach` so the `api()` guard can actually fire instead of silently handing back the previous test's hook.

**Issue 5 — CI step order.** Fixed, and taken further than suggested. `Setup Node` / `Install dependencies` / `Lint` are now hoisted above `Setup Rust`, so lint reports before any Rust toolchain work at all, not merely before `Build WASM`. Step order is now: checkout → Node → install → **Lint** → Rust → cache → wasm-pack → Build WASM → Typecheck → Tests.

**Issue 9 — `allowFreerolls`.** Fixed by adding a test rather than by changing behaviour. `parseFiles` overwrites `pendingAllowFreerollsRef` unconditionally, so a coalesced batch carries only the last upload's flag. That is pre-existing and out of scope to change, but it is now pinned, so a future per-file fix fails loudly here instead of quietly changing which hands get filtered.

### Not fixed

**Issue 6 — redundant `@vitest-environment` docblock.** Declined; the premise is incorrect. The finding states "no sibling test file carries it", but all four React test files do: `ChartTabs.test.tsx`, `DeepDiveCharts.test.tsx`, `HandHistoryCharts.test.tsx`, and `TournamentCharts.test.tsx`. It is redundant with `vite.config.ts` in all five, but removing it from only the new file would make it the odd one out. If it goes, it should go from all five in a separate cleanup.

**Issue 7 — non-lazy `useState` initializer.** Declined. After the Issue 1 fix the initializer is `{ tournaments: 0, handHistories: 0 }` — a constant literal. Wrapping a two-field constant in an arrow function trades one trivial allocation for one closure allocation and is harder to read. The finding already grades it "irrelevant at this scale".

**Issue 8 — index-based worker accessors.** Declined. Resolving by first-posted `type` does not work: the accessors are needed *before* anything is posted (`creates both workers up front` asserts on `instances` with an empty post log), so there is nothing to discriminate on at that point. The coupling is real but it is documented at the accessor, and the hook creating its two workers in a fixed order is a stable property, not an incidental one.

**Observation — `reanalyzePendingRef` cleared before the empty-tournaments return.** Correct, and worth recording, but genuinely pre-existing and unreachable: `App.tsx:117-121` only calls `runAnalysis()` when `tournaments.length > 0`. Tightening it is a behaviour change that does not belong in a lint-fix PR. The new test now characterises current behaviour, so whoever changes it will see that test fail and have to make a deliberate decision — which is the right outcome.

PR #46, opened 2026-07-31
Defer an analysis request made with nothing to analyse

McDic:
Minor follow-up to #45, which surfaced this during review.

> **Stacked on #45.** Base is `fix/lint-gate-and-hooks-hazards-20260731`, not `master` — this touches the same `pumpAnalyze` that #45 converts to a `useCallback`. GitHub will retarget to `master` automatically once #45 merges.

## The problem

`pumpAnalyze` consumed the pending flag *before* checking whether there was anything to analyse:

```ts
if (!worker || analyzeBusyRef.current || !reanalyzePendingRef.current) return
reanalyzePendingRef.current = false
if (tournamentsRef.current.length === 0) return   // flag already gone
```

A pump that did no work still cleared the flag, so the request was dropped rather than deferred. Moving the check above the assignment makes "pending" mean pending until served.

## Scope

**This is a contract fix, not a bug fix, and it is not reachable today.** `App.tsx` only calls `runAnalysis()` when `tournaments.length > 0`, and it re-calls it whenever the set changes, so a dropped request was always reissued moments later. What changes is that the hook no longer depends on a caller-side invariant it does not enforce and cannot see.

## On the test

The new test drives the second pump through the analyze worker's **error path** rather than a second `runAnalysis()`. That is deliberate: a second `runAnalysis()` would re-set the flag, and the test would pass under either ordering — a false-confidence test. Driving the pump from the worker callback is the only way to observe whether the flag survived.

Mutation-tested: restoring the old ordering fails exactly this test and nothing else.

## Verification

| Step | Result |
|---|---|
| `npm run lint` | 0 errors |
| `npx tsc -b` | clean |
| `npm test` | 284 passed (283 + 1) |

## Why this matters more than "unreachable today"

Raised in review, and a better justification than the one above. The current safety is **positional, not structural**: it holds because `App` happens to be the caller, and its effect happens to run after the hook's `tournamentsRef` sync on the same fiber.

A *child* component calling `runAnalysis()` from its own effect does hit the empty-`tournamentsRef` path — child effects flush before parent effects — and on master that request is silently dropped. So this is one refactor away from mattering, rather than merely hypothetical.


McDic:

**Verdict: APPROVE** — no critical or high issues. Findings verified empirically in a throwaway worktree, not by reading alone: the queue suite was run at PR head (13/13) and again with `useAnalysisWorker.ts` reverted to master's ordering (exactly one failure — the new test, `posted` length 0 vs 1).

### The reordering is safe

Every `pumpAnalyze` caller is one-shot and none retries: `runAnalysis` (`:307`), the analyze `onmessage` result branch (`:260`), `onerror` (`:272`), and the init flush (`:277`). The new early return has no side effects, so a caller that hits it does nothing.

- **No double-post.** The flag is still consumed immediately before the only `postMessage`, and `analyzeBusyRef` guards concurrency. Probe: defer → parse → auto-run → result gives exactly 1 post.
- **No spin.** Two successive spontaneous `onerror`s produce 1 post total.
- **No spurious run.** For a stranded flag to trigger an unrequested analysis, some pump must fire with tournaments present — and every in-app pump trigger other than the error path requires an analyse to have been posted, which itself consumed the flag.
- **`reset()` (`:314`) and the unmount cleanup (`:288`) both clear the flag**, so a deferred request cannot survive either.

### The unreachability claim is true — but only positionally

`App.tsx:117-121` is the only `runAnalysis` caller and is guarded by `tournaments.length > 0`. The hook's `tournamentsRef` sync effect (`:146-148`) is registered *before* App's auto-run effect on the same fiber, so the ref is never stale there.

That guarantee is **positional, not structural**. A probe where a *child* component calls `runAnalysis()` from its own effect does hit the empty-`tournamentsRef` path, because child effects flush before parent effects — and on master the request is silently dropped. The invariant this PR hardens is one refactor away from mattering, which is a stronger argument than "unreachable today".

### Findings

1. **[MINOR]** `useAnalysisWorker.ts:174-176` — the comment claims a delivery mechanism that does not exist. It says the request "is served by the next pump, once tournaments have arrived", but nothing pumps the analyse queue on tournament arrival: the parse-result branch calls `pumpParse()` only. Probe: defer → parse delivers a tournament → **0 analyse posts**.
2. **[MINOR]** The change widens one path to a stuck spinner. In exactly the scenario the new test drives — spontaneous analyse-worker `onerror` with a deferred flag and tournaments present — the pump now posts to a worker that just reported an uncaught error and sets `isAnalyzing: true`. If the worker died at load, the spinner never clears. Master's ordering cannot reach this. The same failure shape already exists on master for the mid-run-request-plus-error case.
3. **[NIT]** The test is coupled to `onerror` calling `pumpAnalyze` (`:272`); someone who later decides not to re-pump after a fatal error breaks it for an unrelated reason.
4. **[NIT]** `useAnalysisWorker.queue.test.ts:216` calls the analyse-worker error "the in-app trigger for this path" — generous, since nothing was ever posted to that worker.
5. **[NIT]** No assertion between the parse and the `onerror`, so the test cannot distinguish "served by the onerror pump" from "served on tournament arrival".

### Pre-existing, flagged not requested

`useAnalysisWorker.ts:310-316` — `reset()` clears `analyzeBusyRef` but never terminates or recreates the analyse worker, so an in-flight sim's result still lands and merges into the freshly-reset state (measured: `bankrollResults` returning as `[{initialCapital: 1}]` after a reset meant to clear everything). Also `reset` is exported but no consumer uses it — `App.tsx` never destructures it — so this API has never been exercised against real usage. Not caused by this PR.

### Clean

Security (no I/O, no user-controlled data, no new deps), performance (the new early return is strictly cheaper), naming, nesting, error handling — all unchanged and fine. `eslint` clean on both files.

McDic:
## Response to review — pushed as `7008844`

Comments only, no behaviour change. 13/13 queue tests, lint and `tsc -b` clean.

### Fixed

**Issue 1 — the overclaiming comment.** Fixed, and this was a real error on my part, not a style quibble. I verified it before changing anything: the parse-result branch calls `pumpParse()` and nothing else, so tournament arrival genuinely does not pump the analyse queue. My comment described a delivery path the code does not have.

The comment now says what the code actually does — the flag is preserved for whatever pump comes next — and records why wiring real delivery was *not* the right call here:

> Wiring one into the tournamentsRef sync effect would deliver it, but in App's shape it turns one analysis into two: the arrival pump posts, App's auto-run effect re-sets the flag, and the result triggers an identical re-run. Preserving the flag without wiring delivery is the deliberate stopping point.

That measurement is the reviewer's, and it is the reason this PR stops where it does rather than "completing" the feature.

**Issue 4 — "the in-app trigger".** Fixed. It was generous, for exactly the stated reason: nothing was ever posted to that worker, so realistically `onerror` only fires there on a module or WASM load failure. The comment now says the error is simply the only pump lever left once a second `runAnalysis()` is ruled out, and that it is contrived on purpose.

### Not fixed

**Issue 2 — the widened stuck-spinner path.** Real, and I am leaving it, but it deserves to be recorded rather than waved away. In the spontaneous-`onerror`-with-deferred-flag case the pump now posts to a worker that just reported an uncaught error and sets `isAnalyzing: true`; if that worker died at load, the message goes nowhere and the spinner never clears.

Two reasons not to address it here. It is unreachable from `App.tsx` for the same reason as everything else in this PR, and the same failure *shape* already exists on master for the mid-run-request-plus-error case — so this is a slightly wider door onto a room that was already open. The actual fix is recreating the worker in the error handlers instead of posting to a corpse, which is a genuine behaviour change to error recovery and does not belong in a two-line reordering. Worth its own issue.

**Issue 3 — coupling to `onerror` re-pumping.** Accepted as-is. It is the only remaining pump lever once a second `runAnalysis()` is ruled out as vacuous, so the coupling is not incidental — it is the mechanism under test. The expanded comment now makes that explicit, so anyone who removes the error re-pump will see why this test exists rather than deleting it as noise.

**Issue 5 — no assertion between the parse and the `onerror`.** Declined deliberately, and the review already anticipated this ("arguably you *want* it to survive that refactor"). Adding `expect(posted).toHaveLength(0)` there would pin the *current absence* of an arrival pump. If someone later wires delivery on tournament arrival — a legitimate improvement — that assertion fails for a reason unrelated to what the test is named for. The test should assert that the flag survived, which it does; it should not also assert which pump happened to serve it.

### On the positional-vs-structural point

Not a finding, but the best observation in the review, and I have added it to the PR description. A *child* component calling `runAnalysis()` from its own effect does hit the empty-`tournamentsRef` path, because child effects flush before parent effects. That makes the current safety positional — it holds because `App` happens to be the caller and its effect happens to run after the hook's ref sync — rather than structural. That is a considerably stronger justification for this change than "unreachable today", which is how I originally framed it.

PR #47, opened 2026-07-31
Unify the WASM toolchain setup behind a composite action

McDic:
Follow-up to the discussion on #45. The two workflows each hand-rolled the same four setup steps, and the copies had drifted.

## The drift

| | `test-web.yml` | `deploy-web.yml` |
|---|---|---|
| wasm-pack | `jetli/wasm-pack-action`, pinned `v0.15.0` | `cargo install wasm-pack`, **unpinned** |
| measured | 0s | **77s / 78s** |

Two consequences, one measured and one latent:

**It compiled from source on every deploy** — 77s and 78s on the last two runs. Not fixable by caching, and the logs show why: the cache *hits* and the compile happens anyway.

```
04:33:41  Cache hit for: v0-rust-wasm-build-Linux-x64-db7c195c-deba8a24
04:33:48  Cache restored successfully
04:35:05  Installing /home/runner/.cargo/bin/wasm-pack     ← 77s later
```

Both workflows share the rust-cache key `wasm-build`. `test-web.yml` runs on every push and PR and never `cargo install`s anything, so it wins the save and stores a `~/.cargo/bin` with no wasm-pack in it; the deploy restores that and has to compile.

Confirmed in review: that same deploy run's rust-cache post step printed `Cache up-to-date.` — a full-match restore that saved nothing. So the deploy never had a chance to cache what it compiled.

**Unpinned meant the published artifact was one upstream release away from being built by a different wasm-pack than the PR gate validates.** It resolves to the same v0.15.0 today only because that happens to be the latest (tagged 2026-05-15) — so this is latent, not active. It stops being latent the day rustwasm ships v0.16.0.

## The fix

Both workflows now call `.github/actions/setup-wasm-build`, so the toolchain that gates a PR is *by construction* the toolchain that builds what ships.

The action takes **no inputs**, deliberately. An input for the wasm-pack version would hand a caller the means to drift again, which is the exact failure being consolidated away.

Two things beyond the minimum, called out so they're easy to veto:

1. **The Rust toolchain and cache steps moved in too**, not just wasm-pack. They were duplicated on the same argument, and a silently mismatched `shared-key` degrades caching without failing anything. Happy to narrow this to wasm-pack alone if you'd rather.
2. **Both path filters now list the action.** Without that, a change to the pin would land untested by the very workflow that validates it.

Also updated the `WEB_MIGRATION_PLAN.md` snippet, which still showed `cargo install wasm-pack` — that is where the drifted copy came from.

## Validation caveat

`test-web.yml` exercises the composite action on this PR, so the action itself is validated here.

**`deploy-web.yml` is not** — it only triggers on push to `master`, so its call site cannot run until this merges. The call site is a one-liner of the same form the test workflow just proved, and the resulting step sequence is byte-identical to what deploy ran before apart from the wasm-pack step. If you want certainty before merging, `gh workflow run deploy-web.yml --ref chore/unify-wasm-pack-install-20260731` would exercise it — but that performs a real Pages deploy from this branch, so I have not run it.


McDic:

**Verdict: APPROVE** — no critical or major findings. Everything flagged as risky was verified against real run logs rather than reasoned about.

### Caching semantics — the subtlest risk, and it is clean (verified)

The concern that `workspaces: ". -> target"` might resolve differently inside a composite action is **disproven by direct log comparison**. All three are byte-identical:

| Run | Workspaces | Cache key |
|---|---|---|
| master deploy 30606731013 (before) | `/home/runner/work/pokercraft-local/pokercraft-local` | `v0-rust-wasm-build-Linux-x64-db7c195c-deba8a24` |
| master test-web 30606731028 (before) | same | same |
| PR test-web 30606927670 (inside composite) | same | same |

**The post/save hook ordering does not bite.** The dangerous failure mode would be the nested `rust-cache` post step firing when the *composite* finishes — before `Build WASM` populates `target/` — silently poisoning the cache while staying green. It does not:

```
 9  success  Run tests                        [05:30:04 -> 05:30:19]
16  success  Post Setup WASM build toolchain  [05:30:19 -> 05:30:19]
17  success  Post Setup Node.js
```

End of job, correct LIFO order, and the log inside it reads `Cache up-to-date.` — exactly what master did.

### deploy-web.yml will work on master (verified as far as possible without merging)

Both workflows' step lists were expanded programmatically through the composite and diffed. The `build` job is identical to master apart from the intended wasm-pack swap; `test-web.yml` is exactly identical. Job-level config, workflow-level `permissions`/`concurrency`, and the `build`→`deploy` dependency are unchanged. No `steps.<id>` reference points at a removed step. Checkout precedes the `uses: ./...` in both jobs.

Two further confirmations that de-risk the unexercised deploy path:

- **PATH propagation from inside a composite works** — jetli writes `/home/runner/.cargo/bin/wasm-pack`, and the workflow-level `Build WASM` step found it.
- **The artifact is unchanged.** Output from master's `cargo install`-built wasm-pack vs the PR's musl prebuilt v0.15.0 is size-identical across all four files: `pokercraft_wasm_bg.wasm` 404567, `pokercraft_wasm.js` 26429, `pokercraft_wasm.d.ts` 7229, `pokercraft_wasm_bg.wasm.d.ts` 2545.

### One claim the PR hedged is actually provable

The PR body labelled as inference the claim that deploy's compiled wasm-pack was never cached. It is **provable from the master deploy log**: that run's rust-cache post step printed `Cache up-to-date.`, meaning the restore was a full match and deploy saved nothing. `cargo install wasm-pack` there ran 05:24:59.6 → 05:26:15.6 = **76s**, and installed v0.15.0, so the "same version by coincidence" claim holds too.

### Findings

1. **[MINOR]** Rationale now lives in three places — the action plus both call sites. A PR whose thesis is "duplicated copies drift" ships three prose copies of one decision, two describing the history of a step that no longer exists at that location.
2. **[MINOR]** The docs edit leaves a stale snippet looking fresh. Appendix B is part of a dated (2026-02-15) historical plan; the one line fixed sits beside six still-wrong facts — `branches: [main]`, `crates/wasm-bindings`, the output dir, `node-version: '20'`, `peaceiris/actions-gh-pages@v4`, and the filename `web-deploy.yml`. Separately, `uses: ./.github/actions/setup-wasm-build` is meaningless in a snippet meant as a portable example.
3. **[MINOR]** `scripts/build-wasm.sh` is in neither `paths:` filter. Both workflows execute it; neither triggers on it. This PR's own justification for adding the action applies verbatim.
4. **[MINOR]** Observability regression — the composite collapses three steps into one in the job step list and the jobs API. The per-step timings this PR's argument is built on are no longer readable from the step list, only from expanded logs.
5. **[NIT]** Fourteen header lines restate GitHub's own documentation and will go stale, including a rule for `run:` steps this action does not have.
6. **[NIT]** `jetli/wasm-pack-action@v0.4.0` is on the Node 20 deprecation list and is unmaintained (last release 2023). It now sits between master and the live Pages site.
7. **[NIT]** The `paths:` comment is duplicated across `push` and `pull_request` — unavoidable, since they need separate lists.

### Design decisions, assessed

- **No inputs:** correct. With two callers that must be identical, an input is a drift vector with no current consumer.
- **Folding in toolchain + cache:** correct, and the log proves why — the two copies of `shared-key` had to agree exactly to produce the same key, and a silent mismatch degrades caching without failing anything.
- **Extending both `paths:` filters:** correct. Note the intended consequence that a comment-only edit to `action.yml` on master now triggers a real Pages deploy.

### Clean

Security and supply chain (no new secret handling, token exposure unchanged, `pull_request` from forks gains no new capability), permissions, concurrency, artifact upload, and the `build`→`deploy` dependency.

McDic:
## Response to review — pushed as `c0a67e6`

Four of seven findings fixed, three declined with reasons.

### Fixed

**Issue 1 — rationale in three places.** Fixed, and the finding is exactly right about the irony: a PR arguing that duplicated copies drift shipped three prose copies of one decision, two of them narrating the history of a step that no longer exists where they sat. Both call sites are now a one-line pointer at the action, which is the only place the reasoning lives.

**Issue 2 — the docs edit.** Fixed by **reverting** it rather than refining it, which is the better call. Patching one line inside Appendix B made a stale snippet look maintained while six other facts in the same block stayed wrong, and it pasted a repo-specific `uses: ./.github/actions/setup-wasm-build` into what is meant to be a portable example — incoherent on its own terms.

The snippet is restored to its original form and Appendix B now opens with a banner saying plainly not to copy from it, listing what has drifted, and pointing at the real workflow files. That addresses the actual hazard — someone copying `cargo install wasm-pack` back out of the docs, which is where the drifted copy came from — without pretending to maintain a dated historical document.

**Issue 3 — `scripts/build-wasm.sh` missing from both filters.** Fixed. This one lands squarely on my own argument: I justified adding the action to the filters on the grounds that a change would otherwise land untested by the workflow that validates it, and the script that both workflows actually execute was sitting in the same blind spot. Added to both.

**Issue 5 — the mechanics primer.** Fixed. The generic parts restated GitHub's documentation and would go stale, including a paragraph about the `shell:` requirement for `run:` steps this action does not have — a rule documented for code that does not exist. Trimmed to what is specific to this repo: the checkout dependency (a real constraint on callers), the no-inputs rationale, and the wasm-pack pin.

### Not fixed

**Issue 4 — observability regression.** Accepted, not fixed, because it is inherent to composite actions rather than to this change: collapsing three steps into one is the mechanism, so the only "fix" is not extracting them. Worth stating the cost plainly — the 77s-vs-0s measurement that motivated this PR would now have to come from expanded logs rather than the step list. For three cheap setup steps that is a fair trade; for something where you cared which sub-step was slow, it would be a reason not to extract.

**Issue 6 — `jetli/wasm-pack-action@v0.4.0` unmaintained and on the Node 20 deprecation list.** Real, and worth recording, but out of scope here and not made worse by this PR. It has gated every PR for months; what changes is that it now also sits on the deploy path. Swapping it — `taiki-e/install-action` is the obvious candidate, better maintained and checksum-verifying — is its own change with its own risk, and bundling it into a refactor that is otherwise provably behaviour-neutral would forfeit exactly that property. Worth an issue.

**Issue 7 — duplicated `paths:` comment across `push` and `pull_request`.** Declined; the finding already concedes it is unavoidable. GitHub requires separate `paths:` lists per event, and a comment on only one of the two would be worse.

### On the hedge that turned out to be provable

The review is right that my "inference, not instrumented" caveat about deploy never caching wasm-pack was too cautious: the master deploy log settles it, since its rust-cache post step printed `Cache up-to-date.` — a full-match restore that saved nothing. I have removed the hedge from the PR description. The action's own comment already stated the 76s as fact, which was justified.

### Note on an intended consequence

Flagged in the review and worth making explicit for whoever merges: because `action.yml` is now in `deploy-web.yml`'s `paths:` filter, **a comment-only edit to it on master triggers a real Pages deploy.** That is intended — the action determines how the artifact is built, so it should be able to trigger a rebuild — but it is a live consequence rather than a hypothetical one.

